PR #31, opened 2014-09-22
Adding example to OptionsMerging

sgithens:


jobara:
@amb26 and @sgithens is this something that should still be added to the docs?

amb26:
Well, users are no longer recommended to use fluid.merge() as a standalone function - this is warned halfway down the document. I see we still barely support it, with a couple of dangerous-looking examples in the prefs framework. If we did anything, it would be best to restructure this page to more clearly focus on mergePolicies - which are also something that we don't direct "everyday" users of the framework (if there are any such people) to get involved with, but unfortunately with the current state of the framework most users end up having to meddle with them somewhat once their app gets beyond a certain level of sophistication.

All of this said, the example is basically helpful in that it shows something concrete happening which otherwise might be quite mystifying. We might consider accepting the update someday if any time emerged to polish it up a bit : P

Note that in the "new" framework virtually everything sensitive (members, models, etc.) gets sent into the general "fluid.deferringMergePolicy" since it has become accepted that the action of the generic mergePolicy is faulty because of issues such as  https://issues.fluidproject.org/browse/FLUID-6219

PR #40, opened 2015-02-26
Updated guidelines for test module hash name...

the-t-in-rtf:
In working on my first tests, I discovered that the framework will throw errors unless your test cases are stored in a `modules` hash.  The documentation previously indicated that you should use `testCases`.  I am submitting a pull request to fix this.


amb26:
Thanks very much for this pull request ADTKINS which substantially improves the amenity of our framework!

Could you make sure in future that you do work in branches and issue pull requests from branches (rather than your master) since otherwise coordination problems might arise in the future (for example when you issue further pull requests). In general the idiom is that your master is used to simply track the existing master of the upstream project.

the-t-in-rtf:
Will do.

PR #43, opened 2015-03-16
FLUID-5603: Add the API doc for the text to speech component

jobara:
http://issues.fluidproject.org/browse/FLUID-5603

### DIFF
--- a/src/documents/APIChangesFrom1_5To2_0.md
+++ b/src/documents/APIChangesFrom1_5To2_0.md
@@ -1,5 +1,5 @@
 ---
-title: UI Options 2.0 Migration
+title: API Changes from 1.5 to 2.0
 layout: default
 ---
 
--- a/src/documents/APIChangesFrom1_5To2_0.md
+++ b/src/documents/APIChangesFrom1_5To2_0.md
@@ -3,14 +3,17 @@ title: API Changes from 1.5 to 2.0
 layout: default
 ---
 
-# UI Options 2.0 Migration #
+# API Changes from 1.5 to 2.0 #
 
+## Framework Changes ##
 
-## Namespace Changes ##
+### Preferences Framework ###
+
+#### Namespace Changes ####
 
 Rename "fluid.prefs.enactors" to "fluid.prefs.enactor"
 
-## Component Grade Changes ##
+#### Component Grade Changes ####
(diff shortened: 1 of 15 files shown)


src/documents/APIChangesFrom1_5To2_0.md
```diff
@@ -0,0 +1,133 @@
+---
+title: API Changes from 1.5 to 2.0
+layout: default
+---
+
+# UI Options 2.0 Migration #
```
acheetham: The level one heading should probably be "API Changes from 1.5 to 2.0" – UIO migration things could be grouped under a level two heading?

jobara:
@acheetham I've made adjustments to the headings and brought it more inline with the other migration doc we had fro 1.4 to 1.5.

src/documents/APIChangesFrom1_5To2_0.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Framework Changes ##
+
+### Preferences Framework ###
+
+#### Namespace Changes ####
+
+Rename "fluid.prefs.enactors" to "fluid.prefs.enactor"
+
+#### Component Grade Changes ####
```
acheetham: Are the component grade changes and the model sharing changes restricted to the Prefs Framework, or do they affect all of Infusion, @jobara?

jobara: yes and yes. Basically the grades and model sharing can be used by anything. I think they were actually both in the framework for Infusion 1.5, if I'm not mistaken. However, now we are actually making use of these in the preferences framework.

acheetham: If the new grades and model sharing can be used by anything, then perhaps this information should not be under the "Prefs Framework" heading?

jobara: The thing though is that this was an explicit change made to the prefsFramework that changes the way it is used now. There is documentation in for the model relay system in the infusion docs already. https://github.com/fluid-project/infusion-docs/blob/master/src/documents/ModelRelay.md

jobara:
merged at 6d022603f8cdd4122963ad2a9e44615d5a9f1c30

PR #48, opened 2015-04-13
FLUID-5609: Updated ToC, content, styling

jhung:
@acheetham can you take a look at this?


acheetham:
Looking at the visual appearance of the docs:
- Can we get a highlight on the "current section" link in the upper right corner? i.e. if we're looking at the tutorials, the link "Tutorials" should be highlighted?
- Looking at the Tutorials section table of contents, the various line spacings make it difficult to distinguish one multi-line link from another; the different line spacings seem too similar to each other. Could we reduce the line spacing in a single link?
- Links in the table of contents are bold, but they're not in the designs.
- Could we make the footer text a little bit smaller?

acheetham:
Nice job on highlighting the current page in the table of contents! I notice that for the multi-page "getting started" tutorial, if you're on one of the "sub-pages" the highlight goes away. This is very low priority, but if it happens to be easy to maintain the highlight, it would be nice.

src/layouts/default.html.handlebars
```diff
@@ -54,7 +67,7 @@
                     </div>
 
                 </article>
-            
+
             </div>
 
             <footer class="row infusion-docs-footer">
```
acheetham: Could the link text for the link to fluidproject.org be "the Fluid Project" instead of the URL?

src/documents/index.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: Infusion Documentation
 layout: default
+category: Infusion
 ---
 
-Welcome to the Infusion Documentation.
+This is documentation related to using, configuring, and customizing Infusion. Use the table of contents to navigate the content.
```
acheetham: Perhaps all of these "section introduction" pages could say something more like "This section contains documentation related to..." instead of "This is..."

acheetham:
There's nothing in the designs for this, but could we add a focus style?

acheetham:
Are there plans to add UIO to the docs? (can, open... worms, everywhere...)

acheetham:
@jhung, the CNAME file needs end up in the out folder so that it will be deployed to the gh-pages branch. I think that means it needs to start out somewhere in the src hierarchy.

PR #52, opened 2015-04-16
FLUID-5623: Updated ToC

jhung:


acheetham:
@jhung, could you update this pull request to master? (a couple of other branches have been merged)

site-structure.json
```diff
@@ -114,16 +113,56 @@
...
+            },
+            {
+                "sectionName": "User Interface Options",
+                "pages":
+                [
+                    { "pageName": "User Interface Options", "href": "/tutorial-userInterfaceOptions/UserInterfaceOptions.html" }
+                ]
+            },
+            {
+                "sectionName": "",
```
acheetham: The string templates tutorial does not look like it's in its own section, it looks like it's in the UIO section. I think we should move it to the Prefs Framework section; what do you think?

site-structure.json
```diff
@@ -150,6 +189,7 @@
                     { "pageName": "Uploader", "href": "/UploaderAPI.html" },
                     { "pageName": "Undo", "href": "/to-do/UndoAPI.html" },
                     { "pageName": "Table of Contents", "href": "/to-do/TableOfContentsAPI.html" },
+                    { "pageName": "ChangeApplier API", "href": "/ChangeApplierAPI.html" },
```
acheetham: What do you think about dividing the API ToC up into two sections: Framework and Components? I'd put Framework first, mostly since it's shorter...

jhung: I would prefer to leave it for now rather than make too many changes. Definitely worth looking at later. Thoughts?

acheetham: ok, good point. Let's file a JIRA for that, so we don't forget.

site-structure.json
```diff
@@ -77,7 +76,8 @@
                     { "pageName": "Panels", "href": "/Panels.html" },
                     { "pageName": "Composite Panels", "href": "/CompositePanels.html" },
                     { "pageName": "Conditional Subpanels", "href": "/ConditionalSubpanels.html" },
-                    { "pageName": "Localization in the Preferences Framework", "href": "/LocalizationInThePreferencesFramework.html" }
+                    { "pageName": "Localization in the Preferences Framework", "href": "/LocalizationInThePreferencesFramework.html" },
+                    { "pageName": "Using String Templates", "href": "/tutorial-usingStringTemplates/UsingStringTemplates.html" }
```
acheetham: I think this String Templates tutorial should probably stay in the Tutorials section, @jhung – but in the Prefs Framework section of the tutorials instead of the UIO section.

PR #57, opened 2015-04-20
fix indentation error

malarres:
fix small indentation error that provokes a line to appear as code when it's not


jhung:
I took a look at this and could not find any visible difference between the tab being present and the tab being removed.

Was the tab causing the text to render abnormally? If so, what browser was being used?

malarres:
http://docs.fluidproject.org/infusion/development/UnderstandingInfusionComponents.html was updated since then.
Previously you had a box surrounding the phrase "to customize the component"

amb26:
No longer lays out badly

PR #59, opened 2015-04-20
FLUID-5637: Fixes to high-contrast hover and focus styling

acheetham:


src/static/css/infusion-docs.css
```diff
@@ -351,24 +347,81 @@ div.infusion-docs-note {
     min-width: 13rem;
 }
 
+@media only screen and (max-width: 50rem) {
```
jhung: How was 50rem in this media query decided upon? If trying to match Foundation's values, see the values in this document: http://foundation.zurb.com/docs/media-queries.html

Also it may be useful to add a comment so that we know what this query is for?

acheetham: I have no idea where this came from. I didn't create or edit it. I will investigate. If it's there for mobile purposes, I'll probably remove it.

PR #65, opened 2015-04-24
FLUID-5630: Added Skip links

jhung:


acheetham:
This is not actually working in Safari or Chrome: If I activate one of the 'skip' links, the page moves to the right place, but pressing tab again just continues in the tab order as if the link hadn't been activated.

src/static/css/infusion-docs.css
```diff
@@ -73,6 +73,24 @@ a:focus {
...
+    position: absolute;
+    top: auto;
+    width: 1px;
+}
+
+#skip a:active,
+#skip a:focus {
+    height: auto;
+    position: static;
+    width: auto;
```
acheetham: When the links appear on focus, the whole page shifts down. I wonder if it's possible to do something with the styling so that the links appear over top of the header?

PR #66, opened 2015-04-27
FLUID-5619: TOC fixes, active category styling

jhung:
- An underline now appears below the current active category in the header navigation.
- Added styling to ToC which gives better distinction between TOC items.
- Shrank Footer text to 0.8 rem.

### DIFF
--- a/src/static/css/infusion-docs.css
+++ b/src/static/css/infusion-docs.css
@@ -206,6 +206,10 @@ a:focus {
     border-top: 0.1rem solid #efefef;
 }
 
+.infusion-docs-TOC li:first-child {
+    border-top: none;
+}
+
 .infusion-docs-category {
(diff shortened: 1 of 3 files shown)


src/static/css/infusion-docs.css
```diff
@@ -197,7 +202,8 @@ a:focus {
 }
 
 .infusion-docs-TOC li {
-    margin-bottom: 1rem;
+    padding: 0.6rem 0;
+    border-top: 0.1rem solid #efefef;
```
acheetham: Very subtle point: I wonder if this might work better as a border-bottom. As it is, a top border seems to create a slight separation between the first item in a section and the section header, and a lack of such separation between the last item and the next header, and I wonder if the reverse might come across better.

PR #77, opened 2015-10-18
FLUID-5791:  Comprehensive documentation for the Core API, Promises API, jqUnit API, IoC API and API for use in node.js

amb26:
together with various updates for consistency, as well as fixes for CSS ToC height issue and implementation of "hover to link" functionality as per github's own support for rendering .md files


site-structure.json
```diff
@@ -5,6 +5,16 @@
         "sections":
         [
             {
+                "sectionName": "API",
```
acheetham: I'm not convinced that this API section should be the first thing people see in the nav bar. It seems to me that the more introductory, explanatory pages should remain at the top.

site-structure.json
```diff
@@ -5,6 +5,16 @@
...
+                "pages":
```
acheetham: There are two other API pages further down: ChangeApplier and DOM Binder. Should they be moved into this section?

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+---
+
+# Infusion Core API
+
+This file documents some low-level APIs which are provided by Infusion at the JavaScript level. Note that since
+Infusion's more overarching aim is to enable declarative programming based on JSON structures, by means
+of its [IoC](FrameworkConcepts.md#ioc) system, the documentation in this file doesn't cover very much
+of Infusion's real function. Please consult pages on [Component Grades](#ComponentGrades.md), 
+[Component Configuration Options](#ComponentConfigurationOptions.md), the [Infusion Event System](#InfusionEventSystem.md)
+, etc. for coverage of these wider topics.
```
acheetham: The comma (,) at the start of this line is actually rendering at the start of the line instead of at the end of the previous line

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+category: Infusion
+---
+
+# Infusion Core API
+
+This file documents some low-level APIs which are provided by Infusion at the JavaScript level. Note that since
+Infusion's more overarching aim is to enable declarative programming based on JSON structures, by means
+of its [IoC](FrameworkConcepts.md#ioc) system, the documentation in this file doesn't cover very much
+of Infusion's real function. Please consult pages on [Component Grades](#ComponentGrades.md), 
+[Component Configuration Options](#ComponentConfigurationOptions.md), the [Infusion Event System](#InfusionEventSystem.md)
```
acheetham: All of these links should NOT have the hash (#) in them; that makes them links to anchors on the current page, which is wrong.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+* Returns: `{Boolean}`
+
+Returns `true` if the supplied value is other than `null` or `undefined`.
+
+### fluid.isPrimitive(value)
+
+* `value {Any}`
+* Returns: `{Boolean}`
+
+Returns `true` if the supplied value is of a primitive JavaScript type (recognized: `String`, `Boolean`, `Number`, `Function`, and falsy values)
```
acheetham: Why are 'falsy values' singled out?

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+* Returns: `{Boolean}`
+
+Returns `true` if the supplied object is an Infusion component.
+
+### fluid.isDestroyed(component)
+
+* `component {Component}`
+* Returns: `{Boolean}`
+
+Returns `true` if the supplied Infusion component has been destroyed.
```
acheetham: The word 'destroyed' here might benefit from being a link to the component lifecycle page

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+
+### fluid.hasGrade(options, gradeName)
+
+* `options {Object}`
+* `gradeName {String}`
+* Returns: `{Boolean}`
+
+Returns `true` if the supplied options ([defaults](#fluid-defaults-gradename-options-)) are those of a component which has the supplied grade. These must be options which have
+already been through options merging, since the test is a simpleminded one that simply checks whether the `gradeNames` entry in the options contains the supplied `gradeName`.
```
acheetham: If this only checks the gradeNames entry, then it's a bit misleading to say "if the supplied options are those of a component which has the grade". That phrase is a bit unclear, to begin with: It could be interpreted to imply that if I submit an 'events' option block, then that would successfully match any evented component, since "events" is an option of an evented component. You might want to clarify this to be specifically about the gradeNames, and not the entire options block.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+## Basic implementation atoms
+
+### fluid.identity(arg)
+
+* `arg {Any}`
+* Returns: `{Any}`
+
+### fluid.notImplemented()
+
+A function which raises a failure if executed. This function name can also be used as a placeholder in [Invokers](Invokers.md) and
```
acheetham: "can also be used"... is there any other use case besides the one you describe here? If so, it might be helpful to state it. If not, then "also" is not really appropriate.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+namespaced [Listener](InfusionEventSystem.md) to create abstract (noninstantiable) [Component Grades](ComponentGrades.md).
+
+### fluid.each(source, func)
+
+* `source {Array|Object}` A container to be iterated over (may be null)
+* `func {Function} (element {Any}, index {Number|String}) → Any`. This function will be invoked once on each member of `source` and will receive the member and its key (either an integer for an array source,
```
acheetham: The rendering of this function signature is really not very clear; I think a closing ) is missing? and not sure what the "Any" refers to? Also, the text description should probably use the same words as the parameters (e.g. I think 'member' is referring to 'element', but I wouldn't have to make that assumption if the same word was used).

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+or a string for an object source) 
+* Returns: none
+
+Iterates over a supplied array or hash, invoking a function once for each element. Similar to [`jQuery.each`](http://api.jquery.com/jquery.each/) only the
+arguments to `func` are the right way round and the function does not explode on nonvalues. `fluid.each` on `null` or `undefined` is a no-op.
+
+### fluid.invokeLater(func) 
+
+A standard utility to schedule the invocation of a function after the current function call stack returns. On browsers this defaults to `setTimeout(func, 1)` but in
```
acheetham: All descriptions above have the parameters described after the heading, followed by the textual explanation. Is there a reason for the switch here?

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+Return an empty container as the same type as the argument (either an array or hash). To be wise, the supplied argument should 
+pass the `fluid.isPlainObject` test.
+
+### fluid.copy(tocopy)
+
+* `tocopy {Any}`
+* Returns: `{Any}`
+
+Performs a deep copy (clone) of its argument. This will guard against cloning a circular object by terminating if it reaches a path depth
+greater than `fluid.strategyRecursionBailout`. `fluid.copy` will not copy an object which passes `fluid.isUncopyable`.
```
acheetham: We might want to mention what the default value of fluid.strategyRecursionBailout is

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+* If it is even slightly arrayable (has a numeric `length` property), its elements will be copied into a fresh array
+* If it is not slightly arrayable, it will be placed into a 1-element array
+* If it is undefined, a 0-element array will be returned
+
+### fluid.generate(n, generator[, applyFunc])
+
+Fills an array of given size with copies of a value or result of a function invocation
+* `n {Integer}` The size of the array to be filled
+* `generator {Object|Function}` Either a value to be replicated or function to be called. In case of a function, the signature is `(i {Integer}) → Any`, accepting the index position and returning the generated element.
+* `applyFunc {Boolean}` If `true`, the `generator` value is to be treated as a function
```
acheetham: You should clarify that if applyFunc is false, then generator will just be copied, even if it is a function. And I ask: if generator is a function and applyFunc is false, am I still bound by the signature described above? And does applyFunc have a default value?

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
+---
+title: Core API
+layout: default
+category: Infusion
+---
+
+# Infusion Core API
+
+This file documents some low-level APIs which are provided by Infusion at the JavaScript level. Note that since
```
acheetham: There is an awful lot of info on this page (which is great!). It might be helpful to add a table of contents of at least the sections, so that readers have a sense of what's here, and to help them find what they're probably looking for.

amb26: The ToC widget is freely available for users of the "display preferences" widget installed at the top of the page! And I fear that if a literal ToC was added to the markup, it would end up confusingly duplicating the output from the widget.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+* `fn {Function: (object {Any}, index {Number|String}) → Booleanish }` A predicate function determining whether an element should be
+removed. This accepts the standard signature `(object, index)`. If the predicate returns a "truthy" value, the corresponding object
+will be removed from the source structure.
+* `target {Array|Object}` [optional] A target object of the same type as `source`, which will
+receive any objects removed from it.
+* Returns: `{Array|Object}` Either `target`, containing the removed elements, if it was supplied, or else `source`
+modified by the operation of removing the matched elements.
+
+
+### fluid.getMembers(holder, path)
```
acheetham: The explanation for this function isn't quite clear. I'm not sure how to improve it, since it's still not clear to me what it does :-) "each top-level member": does that mean that the path is relative to the top level? what would the paths be in the resulting object? what does 'path' do if the holder is an array?...

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+Extracts a particular member from each top-level member of a container, returning a new container of the same type.
+* `holder {Array|Object}` The container to be filtered
+* `path {String|Array of String}` An [EL path](FrameworkConcepts.md#el-paths) to be fetched from each top-level member
+* Returns: `{Array|Object}` A container of the same type as `holder`, with the members of `holder` deferenced by `path`.
+
+
+### fluid.filterKeys(toFilter, keys[, exclude])
+
+Accepts an object to be filtered, and a list of keys. Either all keys not present in
+the list are removed, or only keys present in the list are returned. The suppled object *** is not modified *** but will be shallow cloned by this operation.
```
acheetham: I think the explanation of this function – the either/or section – needs some correction, because it seems to be describing the same thing twice...

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+
+
+### fluid.arrayToHash(array)
+    
+Converts an array into an object whose keys are the elements of the array, each with the value `true`
+* `array {Array of String}` The array to be converted to a hash
+* Returns: `{Object}` An object with value `true` for each key taken from a member of `array`
+
+
+### fluid.hashToArray(hash, keyName[, func])
```
acheetham: The description of hashToArray(), including the parameters, needs a bit of clarification. Very not clear... I can't really suggest improvements, since I'm totally confused as to what it does.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+* `array {Array}` The array to be flattened
+* Returns: `{Array}` The flattened array.
+
+
+## Sorting and searching
+
+### fluid.find(source, func[, deflt])
+
+Scan through a list or hash of objects, terminating on the first member which
+matches a predicate function. The return is the return value from the predicate.
```
acheetham: I think the word "matches" here implies an equivalency that is inaccurate; perhaps "satisfies" instead? (same for 'matches' in the description of 'deflt')

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+* `value {Any}` The value to be searched for
+* Returns: `{Boolean|Undefined}` Returns `true` if the value was found in the container
+
+### fluid.keyForValue(tosearch, value) {
+
+Searches through the supplied object for the first value which matches the one supplied.
+
+* `tosearch {Object|Array}` The structure to be searched for the supplied value
+* `value {Any}` The value to be found. This will be compared against the structure's members using `===` equality.
+* Returns: `{String|Integer|Undefined}` The first key whose value matches the one supplied, or `undefined` is no value matches
```
acheetham: Typo: "is" should be "if"

acheetham: Also, it would probably be good to be explicit about the fact that if tosearch is an array, the return value is the index of the found item

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+
+### fluid.stableSort(array, func) 
+
+Applies a stable sorting algorithm to the supplied array and comparator (note that [Array.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) in JavaScript is not specified
+to be stable). The algorithm used will be an insertion sort, which whilst quadratic in time, will perform well
+on small array sizes.
+* `array {Array of Any-Element}` The array to be sorted. This input array will be modified in place.
```
acheetham: The phrase "this input array will be modified in place" should probably be bolded, the way you have with other similar warnings

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+* `func {Function: (a {Any-Element}, b {Any-Element}) → Number}` A comparator returning >0, 0, or <0 on pairs of elements representing their sort order (same contract as [Array.sort comparator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort))
+
+
+### fluid.compareStringLength(ascending)
+
+Produces a comparator function suitable for use with [Array.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) to sort an array of strings by length
```
acheetham: Would the comparator function also be suitable for use with fluid.stableSort()?

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+
+* `ascending {Booleanish}` `true` if a comparator is to be returned which sorts strings in descending order of length
```
acheetham: "ascending true" sorts in *de*scending order? Is that right??

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+
+* `template {String}` A string with embedded tokens of the form `%value`
+* `terms {Object: String → Stringable}` A hash which maps tokens to their required interpolated values   
+
+
+## Handling models and paths
+
+### fluid.get(model, path)
+
+Accesses an element nested within a [model object](FrameworkConcepts.md#model-object] at the supplied (EL) path. `fluid.get` can
```
acheetham: Typo: there's a closing ] at the end of the link instead of )

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+operate various special behaviour in encountering particular kinds of material - for example, any object implementing a function named `resolvePathSegment` will
```
acheetham: This explanation of 'special behaviours' is not terribly clear; for example: "any object implementing a function named resolvePathSegment" - what does this mean? is this 'resolvePathSegment' function something the reader would create? what does it mean for an object to implement the function? ...

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+It still, however, operates the same semantic as `fluid.get` for navigating over undefined values.
+
+* `model {Any}` The model object to be indirected into. This may be `undefined`, etc. without triggering an error.
+* `path {Array of String}` An array of path segments representing representing a chain of properties to be navigated
+* Returns: `{Any}` The result of the indirection of the path into the supplied model
+
+### fluid.set(model, path, newValue)
+
+Writes a new value into a model at a specified path. This is performed by directly modifying the penultimately referenced model object (the one referenced by the first `n-1` 
+path segments in `path`) by writing the value `newValue` to its property named after the final path segment. Other approaches to this task are possible
```
acheetham: The comments from "Other approaches..." to "... could admin that." are not necessary in this documentation.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+path segments in `path`) by writing the value `newValue` to its property named after the final path segment. Other approaches to this task are possible 
+(see [New New Notes on the ChangeApplier](https://wiki.fluidproject.org/display/fluid/New+New+Notes+on+the+ChangeApplier#NewNewNotesontheChangeApplier-Theseveralwaysofactuallyapplyingachange)) but the
+fundamental invariant to be respected is that 
+
+```
+    fluid.get(model, fluid.set(model, path, value), path) === value
+```
+Even a functional programmer could admit that. A slight wrinkle, however, is the behaviour of the special path `""` or `[]`, attempting to write a value at the model 
+root. This API is not capable of this and will fail in this case - since it is hampered at the language level by being unable to rewrite the object reference `model` which exists
+in the scope of its caller. The [ChangeApplier](ChangeApplierAPI.md) handles this situation regularly since under its new semantics, the reference `model` is at a known
```
acheetham: This discussion of the ChangeApplier would be better phrased as an instruction to use the ChangeApplier in this case (if that's what you mean) rather than a conversational comment referencing new vs old semantics, which is irrelevant to API documentation.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+
+
+
+## Logging and error handling
+
+### fluid.log([logLevel, arg1, ... argn])
+
+Log a message to a suitable environmental console (in current implementations, `console.log`). If the first argument to `fluid.log` is
+one of the members of the [`fluid.logLevel`](#fluid-loglevel) structure, this will be taken as the priority of the logged message - else it 
+will default to [`fluid.logLevel.INFO`](#fluid-loglevel). If the logged message priority does not exceed that set by the most recent call to the [`fluid.setLogging`](#fluid-setlogging-loglevel-) function,
```
acheetham: This line seems incomplete – it ends with a comma?

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+
+* `seg1 ... segn {String}` The path segments to be composed
+* Returns: `{String}` The EL path composed by concatenating the path segments.
+
+
+
+
+## Logging and error handling
+
+### fluid.log([logLevel, arg1, ... argn])
```
acheetham: Is it intentional that all of the arguments are optional? even arg1 to argn?

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+Accepts one of the members of the `fluid.logLevel` structure or a `LogLevel` object in general. Returns `true` if
+a message supplied at that log priority would be logged at the current logging level. Clients who
+issue particularly expensive log payload arguments are recommended to guard their logging statements with this
+
+* `testLogLevel {LogLevel}` The log level against which the system's current logging level is to be checked.
+* Returns: `{Boolean}` - `true` if a message supplied at that log priority would be logged at the current logging level.
+
+### fluid.doLog([arg1, ... argn])
+
+The actual implementation to which `fluid.log` messages and other logging generated by the system. Implemented to forward to `console.debug` and `console.log` but may be monkey-patched by
```
acheetham: I think there's something grammatically wrong with this first sentence...

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+a user message. As Infusion evolves to handle a more finely graded variety of dynamic authoring scenarios, this function will be broken up into variants which can signal failures of different severities.
+
+The current system behaviour is to throw a particular subclass of the standard `Error` type, of type `fluid.FluidError`. The behaviour of `fluid.fail` can be customised by contributing handlers to
+the standard event `fluid.failureEvent`
+
+### fluid.failureEvent
+
+Every call to `fluid.fail` triggers a firing of this global event. This is a standard [Infusion Event](InfusionEventSystem.md) with several handlers installed on startup to take on various functions.
+Extra handlers are registered, and the builtin handlers overriden, to customise error handling behaviour in various contexts - for example, when Infusion is running in node.js, is running a 
+[jqUnit test](https://wiki.fluidproject.org/display/docs/Writing+JavaScript+Unit+Tests) ***TODO - obsolete docs for jqUnit*** or is running in [Kettle](https://github.com/fluid-project/kettle) etc.
```
acheetham: Should this TODO be here? perhaps put it in brackets, or something? It's odd, the way it is.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+
+Returns any value held at a particular global path. This may be an object or a function, depending on what has been stored there. 
+
+* `path {String|Array of String}` The global path from which the value is to be fetched
+* Returns: `{Any}` The value that was stored at the path, or `undefined` if there is none. 
+
+### fluid.setGlobalValue(path, value)
+
+Write a value to a particular global path. Users are strongly discouraged from storing any mutable information at these paths, but it is accepted that in certain integration scenarios there can be little alternative. 
+Please choose a properly qualified global path name that identifies your project (perhaps, derived from a [domain name](https://en.wikipedia.org/wiki/Domain_name) or [npm package name](https://docs.npmjs.com/misc/faq#why-no-namespaces)
```
acheetham: The phrase beginning with "(perhaps" is missing a closing )

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+we intend that Infusion global names should be handled in an analogous way. 
+
+* `path {String|Array of String}` The global path at which the value is to be set. 
+* `value {Any}` The value to be written to the global path. This should in general be a function or other immutable value, and an existing value should not be overwritten. This rule can be overlooked in a genuine architectural emergency.
+
+### fluid.invokeGlobalFunction(functionPath, args)
+
+Invokes a function held at a particular global path with the supplied arguments. Equivalent to `fluid.getGlobalValue` followed by `Function.apply(null, args)`.
+
+* `path {String|Array of String}` The global path holding the function which is to be invoked
```
acheetham: This 'path' bullet point should say 'functionPath' to match the signature above

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+* `args {Array|Any}` The array of arguments to be supplied to the function. If `args` is not an array, `fluid.makeArray` will be called on it first. 
+* Returns: {Any} Any return value from the function which was invoked
+
+### fluid.invokeGradedFunction(name, spec)
+
+Invoke a global function by name and named arguments. A courtesy to allow declaratively encoded function calls
+to use named arguments, with names encoded in the defaults for a `fluid.function`, rather than bare arrays.
```
acheetham: This mention of "fluid.function" is all-new, and unexplained. Perhaps somewhere explain what it is and what it's for?

acheetham: Actually, the entire description of this function is a bit vague. I think it assumes a lot of knowledge on the part of the reader – knowledge the reader likely doesn't actually have. Perhaps back up a bit, and provide a bit more explanation?

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
...
+resolve onto a grade derived from `fluid.function`. The defaults record should also contain an entry
+`argumentMap`, a hash of argument names onto indexes.
+* `spec {Object}` A hash where the keys are keys of the functions `argumentMap`, and the values are the argument values to be sent to the function. The keys will be looked
+up in the `argumentMap` and the values resolved into a flat list of arguments.
+* Returns: `{Any}` The return value from the function
+     
+
+## Storing and retrieving defaults
+
+Infusion's component system is organised around blocks of JSON which define [grades](#ComponentGrades.md) - another way of describing a grade is as a component's *** default options ***, which
```
acheetham: This link to ComponentGrades.md shouldn't have a # in it

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,635 @@
+---
+title: Core API
+layout: default
+category: Infusion
+---
+
+# Infusion Core API
```
acheetham: The Markdown files shouldn't have a level one header explicitly. The text in the "title" attribute above will be used as the level one header for the page in the rendered HTML.

src/documents/IoCAPI.md
```diff
@@ -0,0 +1,133 @@
+---
+title: IoC API
+layout: default
+category: Infusion
+---
+
+# Infusion IoC API
```
acheetham: The Markdown files shouldn't have a level one header explicitly. The text in the "title" attribute above will be used as the level one header for the page in the rendered HTML.

src/documents/NodeAPI.md
```diff
@@ -0,0 +1,91 @@
+---
+title: node.js Support and API
+layout: default
+category: Infusion
+---
+
+# Infusion in node.js
```
acheetham: The Markdown files shouldn't have a level one header explicitly. The text in the "title" attribute above will be used as the level one header for the page in the rendered HTML.

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
+---
+title: Promises
+layout: default
+category: Infusion
+---
+
+# Infusion Promises API
```
acheetham: The Markdown files shouldn't have a level one header explicitly. The text in the "title" attribute above will be used as the level one header for the page in the rendered HTML.

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+As commented [below](#commentary-on-requirements), Infusion promises have taken a different set of tradeoffs to many of those elsewhere in the industry. In terms of interoperability,
+Infusion promises are at least universally recognised as a "[foreign thenable](https://github.com/cujojs/when/blob/master/docs/api.md#when)" and hence can be easily adapted into promises of any of the other libraries.
+In terms of promise algorithms, since Infusion promises meet a weaker contract than usual, Infusion promises cannot safely be supplied to the promise
+algorithms of other libraries without adaptation. However, promises from foreign libraries can easily be used within Infusion's algorithms.
+
+## Core Promises API
+
+### fluid.promise()
+
+* Returns: `{Promise}`
```
acheetham: The structure of an Infusion "Promise" object should be documented somewhere. If it already is, this should link to that.

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+object with a boolean entry `isError: true` and a String `message` summarising the reason for the rejection. 
+
+### promise.disposition
+
+The current disposition of the promise may be inspected at any time. This is a `String` value which encodes which, if any, of
+`resolve` or `reject` have been received by the promise. At the fresh construction of the promise, the member `disposition` holds
+the value `undefined`. If the promise has received `resolve`, `disposition` will hold the string `"resolve"`, or if the
+promise has received `reject`, `disposition` will hold the string `"reject"`.
+
+### Commentary on core API
```
acheetham: This information doesn't actually strike me as a commentary on the core api, but rather just a description of promises in Infusion. As such, it would be better suited to the beginning of the file, instead of after the api details. As a reader, I would have gotten more out of the API description above if I already had *this* information in my head, as context.

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+A promise is
+
+* A linked pair of event firers, named `resolve` and `reject`
+* At most one of these two events can be fired, at most one time in total
+* Any listeners registered to either of the events after the point of firing will be able to recover the (unique) fired value at the point of registration
+
+Note that this description does not adequately account for the features of specification-conformant promises more widely used in the industry - since these
+implement chaining and asynchronous behaviours which we do not implement.
+
+### Commentary on requirements
```
acheetham: This type of commentary is not appropriate for this documentation. It should be moved to the wiki, and perhaps, if desired, a link to it could be added to this documentation.

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+Determines whether an object is a promise, for our purposes. Any object with a member `then` of type `Function` passes this test.
+This includes essentially every known variety, including jQuery promises. This test in fact identifies what in other libraries is
+termed a "foreign thenable".
+
+### fluid.toPromise(value)
+
+* `value {Any}` A value to be converted ("hoisted") to a promise
+* Returns: `{Promise}` If the supplied value is already a promise, it is returned unchanged. Otherwise a fresh promise is created with the value as resolution and returned.
+
+Coerces any value to a promise. If it is already a promise, it is returned unchanged. Otherwise, `toPromise` constructs a fresh promise resolving to the supplied value.
```
acheetham: This paragraph basically says the same thing as the description of the return value. This section could be tightened up a bit to avoid this redundancy.

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+* `target {Promise}` A promise which will follow the `source` in its resolution. `target` will receive a call to `then` causing it to resolve when `source` resolves,
+and reject when `source` rejects. 
+
+Chains the resolution methods of one promise (target) so that they follow those of another (source).
+That is, whenever source resolves, target will resolve, or when source rejects, target will reject, with the
+same payloads in each case.
+
+### fluid.promise.map(source, func)
+
+Returns a promise whose resolved value is mapped from the source promise or value by the supplied function. If the input value is not a promise, it will
```
acheetham: Again, there is inconsistency wrt whether the descriptive paragraph comes before the parameter descriptions or after. We should be more consistent.

src/documents/PromisesAPI.md
colinbdclark: I'm not entirely convinced by this, but it's an interesting point. The first question someone who is familiar with Promises but not Infusion will likely ask is "Why did you roll your own Promises library?" @acheetham, what's your rationale for arguing that this section not appropriate for documentation? What would make this information more suitable to you as documentation, assuming that users need to understand the motivations for our system, too.

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+
+* `source {Promise}` A promise which is to be followed in its resolution.
+* `target {Promise}` A promise which will follow the `source` in its resolution. `target` will receive a call to `then` causing it to resolve when `source` resolves,
+and reject when `source` rejects. 
+
+Chains the resolution methods of one promise (target) so that they follow those of another (source).
+That is, whenever source resolves, target will resolve, or when source rejects, target will reject, with the
+same payloads in each case.
+
+### fluid.promise.map(source, func)
```
acheetham: This "mapping" is not very well explained here. Mapped to what? Could you elaborate a bit?

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+be converted first to a promise via `fluid.toPromise()`. If the input promise rejects, its rejection reason will be propagated unmapped. 
+
+* `source {Object|Promise}` An object or promise whose value is to be mapped (if an object, will be converted first to a promise via `fluid.toPromise()`)
+* `func {Function: ({Any})  →  Any}` A function which will map the resolved promise value
+* Returns: `{Promise}` A promise for the resolved mapped value.
+ 
+## Promise algorithms
+
+The only currently implemented promise algorithms are based around a core skeleton operating an array of promises in a linear sequence. These 
+are responsive to an additional element of our promises API, the  [`fluid.accumulateRejectionReason`](#promise.accumulaterejectionreason-error-) "inverse API" described below.
```
acheetham: This link text (fluid.accumulateRejectionReason) is incorrect, and the link itself is broken

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+
+### fluid.promise.sequence(sources[, options])
+
+* `sources {Array of {Any|Promise|Function:(options {Object}) →  Any|Promise}}` An array of sources of values or promises which will be evaluated in sequence.
+* `options {Object}` [optional] A structure of options which will be supplied to function members of `sources`.  
+
+Accepts an array of values, promises, functions returning values or functions returning promises and evaluates them in sequence. Note that a standard name for a "function returning a promise"
```
acheetham: The explanation of what this function does is a bit lacking. The sources are "evaluated" – what does that mean, particularly in the case when they're not Promises or functions?

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+Accepts an array of values, promises, functions returning values or functions returning promises and evaluates them in sequence. Note that a standard name for a "function returning a promise" 
+is a *task* - this implementation can be directly compared to [sequence](https://github.com/cujojs/when/blob/master/docs/api.md#whensequence) in the [when.js library](https://github.com/cujojs/when).
+
+In the case that the source element is a function returning a promise (a task), `fluid.promise.sequence` will ensure that
+at most one of these in "in flight" at a time - that is, the succeeding function will not be invoked
+until the promise at the preceding position has resolved.   
+
+### fluid.promise.fireTransformEvent(event, payload[, options])
+
+* `event {[Event](InfusionEventSystem.md)}` A "pseudoevent" whose listeners are to be treated as sucessive (asynchronous) stages in the process of transforming a payload.
```
acheetham: This link to InfusionEvenSystem.md isn't rendering as a link (I think because it's inside the code styling)

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+* `payload {Any}` The original payload input to the transforming chain.
+* `options {Object}` [optional] A set of additional options to be supplied to each listener in the transform chain. Accepts two special options:
+    * `reverse: {Boolean}` If `true`, the sequence of handlers will be notified in reverse order
+    * `filterNamespaces: {Array of String}` A collection of event [namespaces](InfusionEventSystem.md#namespaced-listeners) to be filtered out of the processing chain for this particular firing 
+
+This is a slightly esoteric but very powerful API. To get a sense of its overall function, it could be compared with the standard [pipeline](https://github.com/cujojs/when/blob/master/docs/api.md#whenpipeline)
+algorithm supplied with [when.js](https://github.com/cujojs/when) - the concept is that an "initial payload" (which may be empty) is successively
+transformed by sequentially, possibly asynchronous, stages of a pipeline of functions. Each function accepts the return value of its predecessor,
```
acheetham: Should "sequentially" be "sequential" here?

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+or the initial `payload` value supplied to `fluid.promise.fireTransformEvent` if it is the first in the chain, and `options` is the last argument to `fluid.promise.fireTransformEvent`.
+
+## Inverse API recognised by promises consumed by sequential algorithms
+
+Both `fluid.promise.sequence` and `fluid.promise.fireTransformEvent` will recognise the following method supplied by the user on any promise
+returned by one of sources in the sequence:
+
+### promise.accumulateRejectionReason(error)
+
+* `error: {Object|Error}` A rejection which has been received from a promise "to the right` of this one in a promise sequence.
```
acheetham: The phrase "to the right" has unmatched quotes, and is too ambiguous. I suggest we use either "before" or "after" as appropriate.

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+Each listener to the "transform event" (we call this a "pseudoevent" precisely because each listener does not receive the *same* argument list
+as with traditional events, but instead receives the returned and resolved value of its precessor) has the following signature:
+
+* `listener {Function:(previousValue {Any}, options {Object}) →  Any|Promise}` where `previousValue` is the resolved return from the previous listener notified in the chain,
+or the initial `payload` value supplied to `fluid.promise.fireTransformEvent` if it is the first in the chain, and `options` is the last argument to `fluid.promise.fireTransformEvent`.
+
+## Inverse API recognised by promises consumed by sequential algorithms
+
+Both `fluid.promise.sequence` and `fluid.promise.fireTransformEvent` will recognise the following method supplied by the user on any promise
+returned by one of sources in the sequence:
```
acheetham: should be "by one of the sources" (the 'the' is missing)

src/documents/PromisesAPI.md
```diff
@@ -0,0 +1,236 @@
...
+Let us imagine the promises in a sequence (array) laid out from left to right, in order of sequential execution. 
+This method is called by a sequential promise algorithm when a promise somewhere in the sequence has rejected. Ordinarily, execution would
+pass directly to the overall rejection handler for the sequence. However, before this happens, the sequence algorithm will pass from
+*** right to left *** from the point of rejection and inspect each of the promises in that section for an `accumulateRejectionReason` implementation.
+
+If an implementation is found, it will be called with the current rejection reason as an argument, and the return value will be used as the new
+rejection reason. The resolution algorithm then continues to the left with this new rejection reason in place, etc. Finally the fully accumulated
+rejection reason will be dispatched to the overall rejection handler.
+
+What familiar exception-handling pattern from synchronous code does this reproduce? It is the *** rethrowing pattern ***, described in the Java context by
```
acheetham: This commentary would be better moved out of the API documentation, perhaps into the wiki somewhere?

src/documents/IoCAPI.md
```diff
@@ -0,0 +1,133 @@
...
+## IoC utility methods
+
+### fluid.pathForComponent(component)
+
+* `component: {Component}` The component whose global path is required
+* Returns: `{Array of String}` The component's global path, as an array of path segments
+
+Each instantiated component in the Infusion system has a stable base path in the system's global [component tree](Contexts.md).
+Records about the component are held in Infusion's [instantiator](#fluid-globalinstantiator) where the path and numerous other
+details can be looked up. This utility method accepts a currently instantiated (not [destroyed](CoreAPI.md#fluid-isdestroyed-component-)
```
acheetham: The "not destroyed" phrase is missing a closing )

src/documents/IoCAPI.md
```diff
@@ -0,0 +1,133 @@
...
+* Returns: `{Object}` An expanded version of the input `material`.
+
+### fluid.getForComponent(component, path)
+
+This is a fairly interesting method. During the instantiation of an IoC component tree, the framework's "focus of attention" moves around the tree
+in a data-driven way. If an [IoC reference](#IoCReferences.md) is seen to a particular path, this will direct attention to instantiate the
+material referenced by the path. `fluid.getForComponent` is the method used internally by the framework in order to ensure that any referenced
+material has actually been instantiated before the results from any IoC reference are filled in in the [expanded](ExpansionOfComponentOptions.md)
+options material. You can use this method too, in order to make sure that any member of a component has been instantiated before you look at it. This
+should not normally be necessary - and the use of this API only makes sense during a construction "fit" - during the "semi-static period", you can
```
acheetham: What is a 'construction "fit"'?

src/documents/IoCAPI.md
```diff
@@ -0,0 +1,133 @@
...
+assume that all members of all visible components are concrete. 
+
+This is an analogue of the API [`fluid.get`](CoreAPI.md#fluid-get-model-path-) for use during the "ginger construction process.
```
acheetham: The reader knows nothing about the ginger construction process, so it's probably not helpful to refer to it.

src/documents/IoCAPI.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Nexus API methods 
+
+### fluid.construct(path, options[, instantiator])
+
+Construct a component with the supplied options at the specified path in the component tree. The parent path of the location must already be a component.
+
+* `path {String|Array of String}` Path where the new component is to be constructed, represented as a string or array of segments
+* `options {Object}` Top-level options supplied to the component - must at the very least include a field `type` holding the component's type. Note that these are expressed
+in the currently hypothetical [post-FLUID-5750](https://issues.fluidproject.org/browse/FLUID-5750) format with `type` alongside the component's options rather than at
```
acheetham: It is very disconcerting to read "currently hypothetical" in API documentation. Perhaps leave out that part, and just describe the format?

src/documents/IoCAPI.md
```diff
@@ -0,0 +1,133 @@
...
+a higher nested level as is currently required in local configuration supplied as [subcomponents](SubcomponentDeclaration.md). 
+* `instantiator {Instantiator}` [optional] The instantiator holding the component to be created - if blank, the [global instantiator](#fluid-globalinstantiator) will be used
+
+### fluid.destroy(path[, instantiator])
+
+Destroys a component held at the specified path. The parent path must represent a component, although the component itself may be nonexistent.
+
+* `path {String|Array of String}` Path where the new component is to be destroyed, represented as a string or array of segments
```
acheetham: Should "new" be in this line?

site-structure.json
```diff
@@ -5,6 +5,16 @@
...
+                [
+                    { "pageName": "Core API", "href": "/CoreAPI.html" },
+                    { "pageName": "Promises API", "href": "/PromisesAPI.html" },
+                    { "pageName": "IoC API", "href": "/IoCAPI.html" },
+                    { "pageName": "API and usage in node.js", "href": "/NodeAPI.html" }
```
acheetham: The link text here and the actual heading on the page should be the same.

src/documents/NodeAPI.md
```diff
@@ -0,0 +1,91 @@
...
+
+Infusion's [core API](CoreAPI.md) and [IoC system](HowToUseInfusionIoC.md) are fully supported in node.js. It is supplied with a
```
acheetham: What is "it" referring to in the phrase "It is supplied..."?

src/documents/NodeAPI.md
```diff
@@ -0,0 +1,91 @@
...
+layout: default
+category: Infusion
+---
+
+# Infusion in node.js
+
+Infusion's [core API](CoreAPI.md) and [IoC system](HowToUseInfusionIoC.md) are fully supported in node.js. It is supplied with a 
+standard [package.json](https://github.com/fluid-project/infusion/blob/master/package.json) file and registered as a module in [npm's registry](https://www.npmjs.com/package/infusion).
+Infusion's global namespace model, as operated through functions such as [`fluid.registerNamespace`](CoreAPI.md#fluid-registernamespace-path-)
+and [`fluid.defaults`](CoreAPI.md#fluid-defaults-gradename-options-) require some care in the node.js environment which makes significant efforts to balkanise
```
acheetham: Suggest you add a comma after fluid.defaults and after environment

src/documents/NodeAPI.md
```diff
@@ -0,0 +1,91 @@
...
+* `foreignRequire {Require}` The instance of `require` to be operated after the module name has been interpolated. If omitted, defaults to Infusion's own `require` (which may not be able to
+see everything you can - since it will very likely be higher up in the module tree than you are)
+* `namespace {String}` If this is supplied, the returned value from `require` will be written into Infusion's global namespace by using the [fluid.setGlobalValue](CoreAPI.md#fluid-setglobalvalue-path-value-) API.
+
+### fluid.module.modules
+
+Holds for public inspection Infusion's records of modules as registered via `fluid.module.register`. This will be a hash of `moduleName` to records of the following form:
+
+* `baseDir {String}` the `baseDir` argument supplied to `fluid.module.register` for this module
+* `require {Require}` the `moduleRequier` argument supplied to `fluid.module.register` for this module
```
acheetham: Typo here: moduleRequier

src/documents/ComponentOptionsAndDefaults.md
```diff
@@ -17,7 +17,7 @@ customize the Inline Edit for their particular needs.
 
 A component creator will typically define default values for as many of the component's options as
 possible, so that the component will work "out of the box," without customization. This is done with
-a call to [fluid.defaults](https://github.com/fluid-project/infusion/blob/infusion-1.5/src/framework/core/js/Fluid.js#L1519-L1539), which registers the component's defaults with the Framework.
+a call to [fluid.defaults](CoreAPI#fluid-defaults-gradename-options-), which registers the component's defaults with the Framework.
```
acheetham: This new link is missing the ".md"

src/documents/FrameworkConcepts.md
```diff
@@ -20,22 +20,25 @@ consist of only the following types: `Array`, `Object`, `String`, `Number` and `
...
 It is possible to replace one object tree with another, but still to maintain a stable reference to the same subproperty `baz`, whoever it happens to be today. 
 This is particularly important in web applications where data claiming to be "your data" can suddenly arrive from anywhere (a JSON feed, some persistence, 
 a particularly aggressive version management system, etc). However it got here, you know it is your data because it is at the right path.
 
-EL expressions within Infusion can be evaluated by the framework utilities `fluid.get()` and `fluid.set()`, and also global functions can be similarly 
-invoked by `fluid.invokeGlobalFunction()`. EL path expressions of this sort are fundamental to Infusion's model-oriented thinking, and the operation of the Infusion [ChangeApplier](ChangeApplier.md).
+EL paths within Infusion can be evaluated (dereferenced with respect to a particular model) by the framework utilities `fluid.get()` and `fluid.set()`, and also global functions can be similarly
```
acheetham: These function references could now be links to the CoreAPI page, couldn't they?

src/documents/FrameworkConcepts.md
```diff
@@ -43,6 +46,8 @@ Events are created automatically by the framework as part of the initialisation
 Infusion Events are so close to the language that they are a suitable replacement/implementation for features found in other frameworks, such as the "delegates" found in Mac OS X's Cocoa environment, 
 or in Microsoft's C#. Since event signatures are completely free (free as in "like a bird", not like either beer or speech), any function can potentially actually be an event listener. It is all a matter of perspective.
 
+A modern parallel to Infusion's events is node.js's [EventEmitter|https://nodejs.org/api/events.html]. The API and intent is very similar.
```
acheetham: Should this be "The API and intent _are_ very similar"?

src/documents/jqUnit.md
```diff
@@ -0,0 +1,305 @@
+---
+title: jqUnit
+layout: default
+category: Infusion
+---
+
+Bundled with infusion is ***jqUnit***, a library modelled on the [xUnit](https://en.wikipedia.org/wiki/XUnit) API style which wraps base functionality provided by the
```
acheetham: The "which" clause here is ambiguous. I suggest leaving out the "modeled on the xunit api style" and changing "which" to "that."

src/documents/jqUnit.md
```diff
@@ -0,0 +1,305 @@
...
+popular [QUnit](https://qunitjs.com/). All of the QUnit base functionality remains available in a jqUnit fixture at the `QUnit` namespace, and you should consult the QUnit documentation
+for details of packaging markup-based fixtures and its own API semantic. However, jqUnit is a complete wrapper and it is possible and recommended to write complete
+test suites without reference to QUnit. jqUnit is based on the 1.x API of QUnit, and not the (currently unreleased) 2.x version of QUnit which is API incompatible with QUnit 1.x.
+
+As well as use in the browser, there is also a (currently unstable) node.js module, [node-jqunit](https://github.com/fluid-project/node-jqUnit) which allows the use of the same
```
acheetham: This phrase "As well as use in the browser" doesn't seem to fit with the rest of the paragraph very well. Is there a type? a word missing?

src/documents/jqUnit.md
```diff
@@ -0,0 +1,305 @@
...
+Starts a group of related tests which will display with the module's name as a prefix. Direct passthrough for [QUnit.module](http://api.qunitjs.com/QUnit.module/)
+
+### jqUnit.test(name, testFunc)
+
+Registers (queues) a synchronous test fixture by providing a callback which will run it. Direct passthrough for [QUnit.test](http://api.qunitjs.com/QUnit.test/). This is equivalent
+to a call to `jqUnit.asyncTest`, where the fixture ends with `jqUnit.start`.
+
+### jqUnit.asyncTest(name, testFunc)
+
+Registers (queues) a synchronous test fixture by providing a callback which will run it. Direct passthrough for [QUnit.asyncTest](http://api.qunitjs.com/QUnit.asyncTest/).
```
acheetham: The sentence describing jqUnit.asyncTest is identical to that describing jqUnit.test - cut and paste error?

src/documents/jqUnit.md
```diff
@@ -0,0 +1,305 @@
...
+
+### jqUnit.start()
+
+Restarts QUnit's progress through its fixtures, which were previously suspended by a call to `jqUnit.stop()`. The idiom is that whilst the system is suspended by `stop`, the
```
acheetham: Would you have any objection to replacing all the "whilst"s with "while"? Not critical, but my brain snags on them...

src/documents/jqUnit.md
```diff
@@ -0,0 +1,305 @@
...
+
+* `value: {Any}` The value to be tested 
+
+Asserts that the supplied value is `true`. Equivalent to [`QUnit.ok(value, message)`](http://api.qunitjs.com/ok/).
+
+### jqUnit.assertFalse(message, value)
+
+* `value: {Any}` The value to be tested 
+
+Asserts that the supplied value is `true`. Equivalent to [`QUnit.ok(!value, message)`](http://api.qunitjs.com/ok/).
```
acheetham: This should say 'false' not 'true'

src/documents/jqUnit.md
```diff
@@ -0,0 +1,305 @@
...
+   }
+```
+* `expected {Object|Array}` A condensed JSON respresentation of a set of assertions to make about a subtree of DOM nodes, or an array of these
+* `node {DOM|Array of DOM|jQuery}` The DOM node to be checked against `expected` - this may also be an array of DOM nodes or a jQuery object.
+
+
+### jqUnit.assertDomEquals(message, expected, actual)
+
+Compare two lists of DOM elements (or jQueries) for being equal by virtue of containing the same DOM elements in the same order. This will
+be achieved by canonicalising the DOM elements onto their ids by means of the canonicalisation function `jqUnit.canonicaliseDom`
```
acheetham: Should jqUnit.canonicaliseDom not be documented on this page?

src/documents/jqUnit.md
```diff
@@ -0,0 +1,305 @@
...
+* `expected {Array of DOM|jQuery}` The expected list of DOM nodes
+* `actual {Array of DOM|jQuery}` The actual list of DOM nodes to be compared with `expected`.
+
+### jqUnit.isVisible(msg, selector)
+
+Asserts that the DOM nodes identified by `selector` are visible, in terms of matching the jQuery [`:hidden`](https://api.jquery.com/hidden-selector/) pseudoselector.
+
+* `selector: {jQueryable}` The selector or other jQueryable identifying the DOM nodes to be tested for visibility
+
+### jqUnit.notVisible(msg, selector)
```
acheetham: Are isVisible and noVisible properly describing match/not match of :hidden? or do they have it backward?

src/static/css/octicons.css
```diff
@@ -0,0 +1,221 @@
...
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  -webkit-user-select: none;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  user-select: none;
+}
+.mega-octicon { font-size: 32px; }
+
+.octicon-alert:before { content: '\f02d'} /*  */
```
acheetham: Given that we only use a single one of these icons, do we really need this entire file? Could we merge the necessary parts into infusion-docs.css?

src/documents/CoreAPI.md
amb26: This seems clear enough to me. The first sentence describes one use, and the second describes others.

src/documents/CoreAPI.md
amb26: Yes, perish the thought that somebody might try to assemble a mental model of what they are doing as a result of reading this documentation : P

src/documents/CoreAPI.md
amb26: Certainly.... a perverse caller can perfectly well call fluid.log() just as they can call console.log()

src/documents/PromisesAPI.md
amb26: This is what is documented in this file - in the lines which directly follow

src/documents/PromisesAPI.md
amb26: The point is that "before" and "after" is ambiguous with respect to whether the sequence algorithm is travelling from left to right or not.

src/documents/IoCAPI.md
amb26: They should at least be made aware that there is something important that they know nothing about. I'll construct some standalone docs on the instantiation process (under the long-standing FLUID-5148) once FLUID-4982 is implemented and we don't have to rewrite it all again.

src/static/css/octicons.css
amb26: I'll leave this up to our designers. I would say that having the original file makes it easier to track provenance, but practices may differ.

src/documents/jqUnit.md
amb26: I was substantially drunk by the time I got to this part of the docs.

colinbdclark: :beers: :wine_glass: :cocktail:

amb26:
Addressed first round of feedback, added further pages on View API, Model Transformations and Function Grades - ready for another look

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,675 @@
...
+Returns `true` if the supplied Infusion component has been [destroyed](ComponentLifecycle.md).
+
+### fluid.hasGrade(options, gradeName)
+
+* `options {Object}`
+* `gradeName {String}`
+* Returns: `{Boolean}`
+
+Returns `true` if the supplied options ([defaults](#fluid-defaults-gradename-options-)) are the top-level options of a component which has the supplied grade. These must be options which
+have already been through options merging, since the test is a simpleminded one that simply checks whether the `gradeNames` entry in the options contains the supplied `gradeName`.
```
acheetham: This description of hasGrade() is still a bit self-contradictory: It says "if the supplied options are the top-level options of...", bu then it says it "simply checks whether the gradeNames entry contains the supplied gradeName". These two things are not the same test.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,675 @@
...
+other environments can be customised - e.g. to `process.nextTick` in node.js. In future, this could be optimised in the browser to not dispatch into the event queue.
+
+### fluid.allocateGuid()
+
+Allocate a string value that will be unique within this Infusion instance (frame or process), and
+globally unique with high probability (50% chance of collision after a million trials). This has the structure `prefix-id` where
+`prefix` is the value of `fluid.fluidInstance` identifying this Infusion instance, and `id` is a increasing integer initialised at
+1 when this infusion instance starts up.
+
+* Returns: `{String}` A string of about a dozen characters that has made moderate efforts to be globally unique.
```
acheetham: For consistency, the description of allocateGuid() should to after the description of the return value.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,675 @@
...
+
+## Transforming and filtering
+
+### fluid.transform(source, [fn1, fn2, ...])
+
+Return a list or hash of objects, transformed by one or more functions. Similar to
+[`jQuery.map`](http://api.jquery.com/jquery.map/), but accepts an arbitrary list of transformation functions and does not explode on null sources. 
+`fluid.transform` on a non-array and non-object (e.g. `null`) is a no-op.
+
+* `source {Array|Object}` The initial container of objects to be transformed. If the source is neither an array nor an object, it will be returned untransformed
```
acheetham: Please pick one format for all these API descriptions – either text description before or after the arguments description – and stick to it, on this page an any others.

src/documents/CoreAPI.md
```diff
@@ -0,0 +1,675 @@
...
+Applies a stable sorting algorithm to the supplied array and comparator (note that [Array.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) in JavaScript is not specified
+to be stable). The algorithm used will be an insertion sort, which whilst quadratic in time, will perform well
+on small array sizes.
+* `array {Array of Any-Element}` The array to be sorted. *** This input array will be modified in place. ***
+* `func {Function: (a {Any-Element}, b {Any-Element}) → Number}` A comparator returning >0, 0, or <0 on pairs of elements representing their sort order (same contract as [Array.sort comparator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort))
+
+
+### fluid.compareStringLength(ascending)
+
+Produces a comparator function suitable for use with [Array.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) or [fluid.stableSort](#fluid-stablesort-array-func-)to sort an array of strings by length
```
acheetham: There's a space missing between the link to stableSort and the word "to" that follows it.

src/documents/PromisesAPI.md
acheetham: @amb26, I guess I'm thinking that it's not appropriate for _API_ documentation. Maybe it's just my own mental model; to me, "API documentation" is _reference_ material. You turn to it to look up the specifics of the API, not to develop a deeper understanding of the design of the functionality. I do believe it's valuable information, but its tone, its purpose... that seems inconsistent with the otherwise brief, direct assistance offered by this API page.

But I could be convinced otherwise by good arguments :-)

src/documents/ViewAPI.md
```diff
@@ -0,0 +1,174 @@
...
+* `text {String}` The string to be entered into the live region - this will be announced "soon" depending on the "politeness setting" attached to the `aria-live` attribute of the region (default: `"polite"`)
+* `options {AriaLabellerOptions}` [optional] A set of options for configuring the behaviour of the labeller. These options are described in the [table below](#structure-of-arialabelleroptions-).
+* Returns: `{Component}` An Infusion [view component](ComponentConfigurationOptions.md#view-components) managing the interaction with the label. ***Note:*** disposing of this component is the responsibility of the caller. If you have destroyed
+the markup to which this component is attached, you must destroy the component using its `destroy()` method or it will leak. Successive calls to `fluid.updateAriaLabel` for the same DOM node will return the same component. 
+
+#### Structure of `AriaLabellerOptions`
+ 
+| Name | Type | Description | Default |
+|------|------|-------------|---------|
+|`dynamicLabel`| `Boolean`| Whether or not an ARIA  live region should be created an associated with the element, as well as the use of `aria-label`| `false` |
```
acheetham: Typo: "created an associated" instead of "created and associated"

src/documents/ViewAPI.md
```diff
@@ -0,0 +1,174 @@
...
+|`liveRegionMarkup`| `String` | The markup to use to create the live region (only used if `dynamicLabel` is `true`).|`"<div class=\"liveRegion fl-offScreen-hidden\" aria-live=\"polite\"></div>"`|
+|`liveRegionId`| `String` | The ID to assign to the live region (only used if `dynamicLabel` is true). |`"fluid-ariaLabeller-liveRegion"`|
+|text| `String` | The text to place in the live region (only used if `dynamicLabel` is true).|none|
+
+Note that as well as forming the 3rd argument to `fluid.updateAriaLabel`, these options also form the [component options](ComponentOptionsAndDefaults.md) to the associated component, which is 
+an Infusion `fluid.viewComponent` of type `fluid.ariaLabeller`.
+
+## Managing Focus and Blur
+
+Infusion includes two "major utilities" for managing focus interactions, [`fluid.globalDismissal`](#fluid-globaldismissal-nodes-dismissfunc-) and [`fluid.deadMansBlur`](#fluid-deadmansblur-control-options-) for managing focus interactions, as well as a few minor utilities.
```
acheetham: "for managing focus interactions" is duplicated in this sentence.

src/documents/ViewAPI.md
```diff
@@ -0,0 +1,174 @@
...
+
+### fluid.globalDismissal(nodes[, dismissFunc])
+
+* `nodes {Object: String  → DomElement|jQuery}` A free hash of names onto nodes that a click ***outside*** of will trigger `dismissFunc`
+* `dismissFunc {Function: DomEvent  →  None}|Undefined` A function which will be executed when a "dismissal click" is received in the document. If this argument is not supplied, any existing dismissal function will be disarmed.   
+    
+Used in the stereotypical situation where a click event (morally, one which shifts focus outside a dialog or other region of interaction - but only actual `click` events
+are supported by this function) should have the effect of "dismissing` the dialog or interaction region. This accepts a free hash of nodes designating the interaction
```
acheetham: The single quote at the end of 'dismissing' should be a double quote.

src/documents/ViewAPI.md
```diff
@@ -0,0 +1,174 @@
...
+region and an optional "dismissal function". If `dismissFunc` is set, this "arms" the dismissal system, such that when a click
+is received OUTSIDE any of the DOM node hierarchy covered by `nodes`, the dismissal function will be executed, and supplied with the DOM event
+triggered by the click event. `fluid.globalDismissal` is a `one shot deal` - if the `dismissFunc` is executed, the system is disarmed and will need
```
acheetham: The quotes around "one shot deal" should probably be double quotes (single quotes result in code-styling).

src/documents/ViewAPI.md
```diff
@@ -0,0 +1,174 @@
...
+* `control {jQueryable}` The control for which a blur interaction is to be managed.
+* `options {DeadMansBlurOptions}` A set of options managing the interaction, described in [table below](#structure-of-deadmansbluroptions-).
+* Returns: `{Component}` A component managing the interaction
+
+This utility is designed to compensate for the fact that delivery of blur events in browsers has always been extremely unreliable. Sometimes blur events
+are delivered simply because focus has moved to another control within the same interaction region, they are not delivered at all, or in some even more
+perverse cases (especially on Internet Explorer) are delivered *after* the focus event which morally causes them. 
+
+This utility uses the following strategy: on receiving a standard `blur` event on the `control`, a timer is started which waits for a short period of
+time (options.delay, defaults to 150ms) to discover whether the reason for the blur interaction is that either a focus or click is being serviced on a nominated
```
acheetham: "options.delay" should probably be code styled here

src/documents/ViewAPI.md
```diff
@@ -0,0 +1,174 @@
...
+require reimplementation (e.g. that its handlers can't be removed from the target elements). It should be considered as an unstable API - however it is still being found
+broadly useful.
+
+#### Structure of `DeadMansBlurOptions`
+ 
+| Name | Type | Description | Default |
+|------|------|-------------|---------|
+|`exclusions`|<code>Object: String  → DomElement&#124;jQuery</code>| A free hash of names onto nodes that a click ***outside*** of will trigger `handler`, *** OR *** that a focus event received within the time window of `delay` from a blur event on `control` will *** cancel *** notification of `handler`| none |
+|`handler`|`Function: (jQueryable)  →  None`| A function which will be invoked when the component has determined that a meaningful blur has been triggered on `control`. It will be invoked with `control` as the argument.| none |
+|`delay`|`Number`| The interval of time that the component will wait after receiving a `blur` event on `control` in order to determine that the reason is that another element within `exclusions` is being focused.|150|
```
acheetham: I know you mentioned above that the delay is milliseconds, but it would probably be a good idea to add that info to this description, too, eg. "The interval of time (in milliseconds) that ..."

src/documents/ModelTransformationAPI.md
```diff
@@ -0,0 +1,84 @@
+---
+title: Model Transformation API
+layout: default
+category: Infusion
+---
+
+The *** model transformation *** framework is a core part of Infusion and allows you to transform change an input [model](FrameworkConcepts.md#model-objects) (JSON structure) based on a set of rules.
```
acheetham: This opening sentence should probably say "transform" or "change", but not both ("allows you to transform change an input...")

src/documents/ModelTransformationAPI.md
```diff
@@ -0,0 +1,84 @@
...
+Any model transformation can be supplied as part of a [model relay](ModelRelay.md#explicit-model-relay-style) rule registered in a [fluid.modelComponent](ComponentConfigurationOptions.md#model-components)'s `modelRelay` 
+block - the system will then constantly operate the transformation rule whenever there is a model change to keep the linked model in sync. The user may also operate model transformation rules manually by use of the
+`fluid.model.transformWithRules` API.
+
+The best documentation for the currently available set of transformation functions is on the [GPII wiki](https://wiki.gpii.net/index.php/Architecture_-_Available_transformation_functions) but will
+in time be migrated here.
+
+## fluid.model.transformWithRules(source, rules[, options])
+
+* `model {Object}` The input model to transform - this will not be modified
```
acheetham: The signature above uses "source" as the first parameter, while the description calls it "model"

src/documents/ModelTransformationAPI.md
```diff
@@ -0,0 +1,84 @@
...
+* `rules {Transform}` A transformation rules object containing instructions on how to transform the model (see [below](#structure-of-a-transformation-document)) for more information)
+* `options {Object}` A set of options governing the kind of transformation to be operated. At present this may contain the values:
+    * `isomorphic {Boolean}` If `true` indicating that the output model is to be governed by the same schema found in the input model, or
```
acheetham: Is it intentional that the last word of the description of "isomorphic" is "or"? This implies that the options can contain either isomorphic or flatSchema. Is that correct?

src/documents/ModelTransformationAPI.md
```diff
@@ -0,0 +1,84 @@
...
+A `{SingleTransform}` record is as follows:
+
+```json
+{
+    type: <transform-type>
+    ...
+}
+```
+
+where `transform-type` is the name of a registered transformation function. This will be a [function grade] derived from the base grade `fluid.transformFunction`. Each transformation function can accept arbitrary additional top-level
```
acheetham: This link to "function grade" is missing the target

src/documents/FunctionGrades.md
```diff
@@ -0,0 +1,66 @@
...
+
+Most [grades](ComponentGrades.md) represent Infusion components - these are derived from the base grade `fluid.component`. However, some grades describe plain JavaScript functions - these are derived from the base grade `fluid.function`.
+The purpose of the Fluid API call [fluid.defaults](CoreAPI.md#fluid-defaults-gradename-options-) could be understood as providing *metadata* about some element of the system. In the case of a full component grade, this metadata is sufficient
+to allow the framework to automatically construct the creator function for the component. In the case of a function grade which describes an already existing function with a global name, the metadata supplies hints to the user about
+how to call the function and its purpose.
+
+## Registering a global function
+
+A global function is registered within infusion at a stable place in its global namespace by working with the core API functions [fluid.registerNamespace](CoreAPI.md#fluid-registernamespace-path-) and 
+[fluid.setGlobalValue](CoreAPI.md#fluid-setglobalvalue-path-value-) - in practice the latter is rarely used, in favour of directly setting members on namespaces objects.
```
acheetham: Should the last two words here be "namespaced object" (i.e. instead of "namespaces")?

acheetham:
@amb26, this is looking good; just a few remaining comments

src/documents/ModelTransformationAPI.md
amb26: Yes, that's right

src/documents/CoreAPI.md
amb26: I'm not sure how this could be clearer. The test which is described, is the one which is actually performed - and the likely purpose of the test is to determine the condition in the first sentence. If the condition is met, then the test will pass, but not necessarily vice versa. Not all tests are necessary and sufficient tests for the criterion tested.

src/documents/PromisesAPI.md
amb26: For some reason I can't see this comment anywhere in the github interface. Here is the argument:

Given that the API structure, both in name and signature, of our promises API, is identical to that of other 
libraries widely available, but at the same time the semantic and behaviour is quite different, it would be 
actively misleading to confine our mission in this "reference material" to simply describing these minutiae 
of the API form. A user who simply takes the form and description of our API and tries to use it, with the 
expectations that they have brought from some other Promises API, without understanding the differences in 
philosophy and intention lying behind our API, would run into serious problems.

amb26:
Thanks, @acheetham - comments addressed for this round

PR #83, opened 2016-02-04
FLUID-5821: Added description of newly specified behaviour for DOM binder mismatches

amb26:
as implemented in FLUID-5821


jobara:
@amb26 could you please fix the NOTE at the top of DOMBinderAPI.md. It mixes markup and markdown syntax.

amb26:
Hi @jobara - this is the style that is used in all other Note entries - I don't think we should fix it in a way that is inconsistent with the others. I suggest we consult with the team responsible for styling the docs to discover why it has been written this way.

jobara:
@amb26 from what I recall the HTML is used because of the required class name, and that it isn't possible to add classes in markdown.

The markdown syntax within HTML doesn't work in github rendered markdown files, but it does seem to work in the docpad generated ones.

jobara:
merged at 94207308ed8604cce0e9d40b56aa2d411457ba5d

PR #84, opened 2016-03-02
FLUID-5695: Documentation updates for model relay improvements

amb26:
Matching https://github.com/fluid-project/infusion/pull/671


amb26:
Work is not yet finished - need to document priorities for relay, priorityHolder, and give an example of user source tracking

amb26:
Now ready for review

src/documents/ChangeApplierAPI.md
```diff
@@ -60,14 +60,90 @@ A `<modelPathReference>` has the form:
...
+
 #### Model Listener Declaration ####
 
-A model listener declaration block has exactly the same form and meaning as any of the record types supported by [Invokers](Invokers.md) and [Listeners](EventInjectionAndBoiling.md#listener-boiling) -
-including the one-string compact syntax documented with [Invokers](Invokers.md#compact-format), and the use of [Priorities](Priorities.md). There are two extra features that are supported in
-model listener blocks that are not supported in standard listener declarations. These are the special context name `change`, and the ability to filter a change based on its _source_.
+A model listener declaration block has the same form and meaning as any of the record types supported by [Invokers](Invokers.md) and [Listeners](EventInjectionAndBoiling.md#listener-boiling) -
+including the one-string compact syntax documented with [Invokers](Invokers.md#compact-format), and the use of [Priorities](Priorities.md). Like standard event listener entries, model
+listeners can provide a `namespace` entry. Just one model listener with a particular namespace will be registered on a particular ChangeApplier.
```
cindyli: Just curious, since the `namespace` entry can be defined at two places: 1. as a `namespace` listener option; 2. the left hand side of the listener declaration. When both are provided, which one would take precedence? An example to clarify what I mean:

```javascript
modelListeners: {
    namespace1: {
        namespace: "namespace2",
        path: "oneModelPath",
        funcName: "fluid.listeners.func",
    }
};
```

In this example, the `path` option is supplied, so the `namespace1` is treated as the model listener namespace. Is the namespace of this listener `namespace1` or `namespace2`?

amb26: Very good point, @cindyli  - I've added an update to make it clear that the `namespace` entry within the record will take priority if it exists.

src/documents/ChangeApplierAPI.md
```diff
@@ -60,14 +60,90 @@ A `<modelPathReference>` has the form:
...
+        <tr>
+            <td><code>String</code></td>
+            <td>A <code>&lt;shortModelPathReference&gt;</code></td>
+            <td>
+                See examples in section [Model Path References](#model-path-references)
+            </td>
+        </tr>
+        <tr>
+            <td><code>Object</code></td>
+            <td>A `<modelPathRecord>`, including members <ul>
```
cindyli: In the context of being wrapped in plain html, the use of \` does not convert the quoted string to code format. Please replace it with <code>.

src/documents/ChangeApplierAPI.md
```diff
@@ -60,14 +60,90 @@ A `<modelPathReference>` has the form:
...
+                </ul>
+            </td>
+        </tr>
+        <tr>
+            <td><code>Array</code></td>
+            <td>An array each of whose element may either be a String holding a <code>&lt;shortModelPathReference&gt;</code> or a <code>&lt;modelPathRecord&gt;</code>. The listener will be notified when <it>any</it> of these
+            paths receive changes. Below is further information on <a href="#matching-on-multiple-paths-in-a-single-model-listener-declaration"><code>matching on multiple paths</code></a></td>
+            <td><code>[
+                    "position", {
+                        segs: [["windowHolders"], "{that}.options.ourWindow"]
```
cindyli: Why is `"windowHolders"` in an array? Is the resolved model path different from the form without having it in an array - `segs: ["windowHolders", "{that}.options.ourWindow"]`?

src/documents/ChangeApplierAPI.md
```diff
@@ -60,14 +60,90 @@ A `<modelPathReference>` has the form:
...
+            </td>
+        </tr>
+    </tbody>
+</table>
+
+#### Matching on multiple paths in a single model listener declaration
+
+When the `path` member of a model listener declaration holds an `Array`, the listener will be notified when <it>any</it> of these paths receive changes. 
 
+Note that a listener which specifies references to multiple component targets in such a list will just receive one notification per component of its changes. A listener which supplies an
```
cindyli: Can you clarify "just receive one notification per component of its changes". 

Just a guess, does it mean when one listener is listening to multiple model paths like "x" and "x.y". When a change is made on "x.y" that trigger model changes on both "x" and "x.y", the listener only receives one notification rather than two?

src/documents/ChangeApplierAPI.md
```diff
@@ -60,14 +60,90 @@ A `<modelPathReference>` has the form:
...
+array of more than one element in <code>path</code> will not be able to make use of either the special context [`change`](#the-special-context-change-) ***or*** the possibility of using the wildcard character
+`*` in the final path segment. Note that elements of <code>segs</code> may themselves consist of IoC references resolving to configuration in the tree (although they may not hold references to model material - they are evaluated just once when the component constructs).
 
 #### The special context `change` ####
 An extra [context name](Contexts.md) is available in a model listener block by the name of `change`. This is bound to the particular change event which triggered this listener. This context behaves as an object with the following fields:
```
cindyli: Shall `triggered` be `triggers`?

src/documents/ChangeApplierAPI.md
```diff
@@ -248,7 +367,7 @@ at the time the component is constructed. Note that programmatically attached li
 The listener is notified after the change (or set of coordinated changes) has already been applied to the model - it is too late to affect this process and so this event is not _preventable_. The signature for these listeners is
```
cindyli: A new line is missing btw this paragraph and the paragraph in front probably because the previous paragraph is using html `<div>`. Try adding a `<br />` at the end.

I'm also confused with the statement of "it is too late to affect this process". What process is too late to be affected? The model change process? Are listeners attached via the declarative way able to stop or prevent the model change by returning a false or something?

cindyli: After a bit more thinking, doe "it is too late to affect this process" mean `excludeSource` and `includeSource` won't take effect for these listeners?

src/documents/ChangeApplierAPI.md
amb26: Thanks, fixed

src/documents/ChangeApplierAPI.md
amb26: An odd anomaly - yes, either form works equally well. To avoid confusion I've removed the extra level of arrays.

src/documents/ChangeApplierAPI.md
amb26: I've clarified by incorporating material from your example.

src/documents/ChangeApplierAPI.md
amb26: triggered is fine. The change event precedes the notification of the listener.

src/documents/ChangeApplierAPI.md
amb26: I didn't suspect that this seemingly harmless aside could cause such confusion - I've simply removed it :) The listeners are all perfectly normal and behave just the way you expect, both declaratively and programmatically. All I was trying to do was link these semantics to the ones for different kinds of events - and pointing out that model events behave like "standard" events rather than "preventable" ones.

src/documents/ChangeApplierAPI.md
```diff
@@ -60,14 +60,91 @@ A `<modelPathReference>` has the form:
...
+A model listener declaration block has the same form and meaning as any of the record types supported by [Invokers](Invokers.md) and [Listeners](EventInjectionAndBoiling.md#listener-boiling) -
+including the one-string compact syntax documented with [Invokers](Invokers.md#compact-format), and the use of [Priorities](Priorities.md). Like standard event listener entries, model
+listeners can provide a `namespace` entry. Just one model listener with a particular namespace will be registered on a particular ChangeApplier. 
+
+A model listener declaration block includes three extra features beyond those found in ordinary event listeners. Firstly is the possibility of including a member [`path`](#the-path-entry-in-a-model-listener-declaration) of type `<modelPathReference>` (a `String` or `Object`), which can hold
+a more complex path specification for the the model listener match than can be encoded in the single string in a `<shortModelPathReference>`, secondly the ability to filter a change based on its [_source_](#source-tracking-and-filtering-in-model-listener-blocks), using the members
```
cindyli: Typo of "the the". Please remove one.

src/documents/ChangeApplierAPI.md
```diff
@@ -60,14 +60,91 @@ A `<modelPathReference>` has the form:
...
+            <th>Meaning</th>
+            <th>Examples</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td><code>String</code></td>
+            <td>A <code>&lt;shortModelPathReference&gt;</code></td>
+            <td>
+                See examples in section [Model Path References](#model-path-references)
```
cindyli: The markdown syntax does not take effect when being placed in a html block. Need to use html instead, such as

`<a href="#model-path-references">Model Path References</a>`

src/documents/ChangeApplierAPI.md
```diff
@@ -60,14 +60,91 @@ A `<modelPathReference>` has the form:
...
+            </td>
+        </tr>
+        <tr>
+            <td><code>Object</code></td>
+            <td>A <code>&lt;modelPathRecord&gt;</code>, including members <ul>
+                <li><code>segs</code> (<code>Array</code> - required) and</li> 
+                <li><code>context</code> (<code>String</code> - optional) which encodes the model path to be matched.</li>
+                </ul>
+                </td>
+            <td>These examples encode exactly the same path expressions, in the same order as in the section [Model Path References](#model-path-references)
```
cindyli: Same as the comment above: replace the markdown with pure html.

src/documents/ChangeApplierAPI.md
```diff
@@ -235,20 +355,23 @@ It is not currently possible to supply more than one wildcard segment per path r
...
+The programmatic style for registering interest in model changes uses an API exposed by the ChangeApplier on its member `modelChanged` that is very similar to that exposed by a standard [Infusion Event](InfusionEventSystem.md) - 
+the difference is that the `addListener` method accepts an extra 1st argument, `spec` - an `Object` which holds the same model path reference in `path` or `segs` documented in the previous section on declarative binding:
 
 ```javascript
-applier.modelChanged.addListener(pathSpec, listener, namespace)
+applier.modelChanged.addListener(spec, listener, namespace)
 applier.modelChanged.removeListener(listener)
 ```
 
+`spec` may also include the standard members `namespace` and `priority` seen in the declarative record.
```
cindyli: When manually calling `applier.modelChanged.addListener()`, the `namespace` can be supplied via: 1. as a property of the `spec` argument; 2. the `namespace` argument. If both are provided, which one will take priority?

src/documents/ChangeApplierAPI.md
```diff
@@ -272,8 +395,13 @@ function listener(value, oldValue, pathSegs, changeRequests)
             <td>An array of <code>String</code> path segments holding the path at which <code>value</code> and <code>oldValue</code> are/were held</td>
         </tr>
         <tr>
-            <td><code>changeRequests</code></td>
-            <td>A single <code>ChangeRequest</code> object or an array of them (see below) which were responsible for this change (may be empty)</td>
+            <td><code>changeRequest</code></td>
+            <td>May contain a single <code>ChangeRequest</code> object which was responsible for this change, but will often be empty. This signature element is not a stable API</td>
```
cindyli: Probably add a link that points to the doc in the next section where `ChangeRequest` objects are explained.

cindyli: Does the statement `This signature element is not a stable API` mean the value of this parameter might be empty when it actually is not empty, or this parameter is not encouraged to use?

src/documents/ChangeApplierAPI.md
```diff
@@ -272,8 +395,13 @@ function listener(value, oldValue, pathSegs, changeRequests)
...
+        </tr>
+        <tr>
+            <td><code>transaction</code></td>
+            <td>May contain a <code>Transaction</code> object which this change was bound to. Primarily useful for the <code>source</code> member which can be used to manually
```
cindyli: Probably add a link to the part where Transaction objects are explained.

src/documents/ChangeApplierAPI.md
amb26: It should not be used, in general. Sometimes there is no single ChangeRequest object responsible for this change.

src/documents/ChangeApplierAPI.md
```diff
@@ -284,13 +412,14 @@ Users will in most cases only be interested in the first argument in this signat
 
 ### Declarative style for triggering a change###
 
-The declarative style for triggering model changes involve a kind of IoC record (a _change record_) that is supported in various places in component configuration, in particular as part of the definition of both [Invokers](Invokers.md) and [Listeners](InfusionEventSystem.md) of an IoC-configured component.
-This style of record is recognised by its use of the special member `changePath` which determines which path in which component model will receive the change.
+The declarative style for triggering model changes involve a kind of IoC record (a _change record_) that is supported in various places in component configuration,
```
cindyli: Shall "involve" be "involves"?

src/documents/ChangeApplierAPI.md
```diff
@@ -378,26 +517,35 @@ applier.change(path, value, type)
...
 </table>
 
 The semantics and values are exactly the same as described in the section on declarative triggering above - with the difference that IoC references may not be supplied for `path`.
 
 ```javascript
-applier.fireChangeRequest(changeRequest)
+applier.fireChangeRequest(<changeRequest>)
 ```
 
-where a **changeRequest** is an object holding the above named parameters in named fields - e.g. `{path: "modelPath", value: "newValue"}` `change` and `fireChangeRequest` reach exactly the same implementation - the only difference is in the packaging of the arguments. For `change` they are spread out as a sequence of 3 arguments, whereas for `fireChangeRequest`, they are packaged up as named fields (`path`, `value` and `type`) of a plain JavaScript object. Such an object is called a **"ChangeRequest"** and is a convenient package for these requests to pass around in in an event pipeline within the framework.
```
cindyli: The last sentence has a typo "in in" (double **in**). Please remove one.

src/documents/CoreAPI.md
```diff
@@ -37,13 +37,16 @@ approach taken from an earlier version of jQuery - detecting whether the `toStri
...
+### fluid.isPlainObject(totest[, strict])
 
 * `totest {Any}`
+* `strict {Boolean}` [optional] If `true`, only accepts plain Objects
 * Returns: `{Boolean}`
 
 Returns `true` if the supplied object is a plain JSON-forming container - that is, it is either a plain Object
-or a plain Array. 
+or a plain Array. If the `strict` argument holds `true`, only returns `true` if the supplied object is a plain Object.
+This method will accept as plain even objects coming from other contexts (e.g. iframes or V8 contexts) if they were
```
cindyli: Typo with "plain even objects"? I'm thinking this line either needs to be clarified or should be removed since why this method cares about what context the given object comes from?

src/documents/CoreAPI.md
```diff
@@ -146,6 +149,14 @@ globally unique with high probability (50% chance of collision after a million t
 `prefix` is the value of `fluid.fluidInstance` identifying this Infusion instance, and `id` is a increasing integer initialised at
 1 when this infusion instance starts up.
 
+### fluid.add(a, b)
```
cindyli: The implementation of `fluid.add()` doesn't check whether `a` and `b` are numbers. Is it callers' responsibility to make sure they are numbers? Otherwise this method is also good for string concatenations etc. :)

src/documents/CoreAPI.md
```diff
@@ -345,6 +367,29 @@ var CATTyears = fluid.hashToArray(hash, "species", function (newElement, oldElem
...
 
+### fluid.freezeRecursive(tofreeze)
+
+* `tofreeze {Array|Object}` The object to be frozen recursively
+* Returns: `{Array|Object}` The supplied argument, recursively frozen
+
+Calls [`Object.freeze`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) at each level of containment of the supplied object.
+This function is a no-op if a primitive value is supplied.
+
+### fluid.pushArray(holder, member, topush)
```
cindyli: I was playing with this interesting function in a debugger console for my little curiosity, I noticed, if supplied `member` value points to an existing object path, and the value of that leaf is not an array, the error `TypeError: array.push is not a function` occurs. If the leaf value is an array, `topush` will be pushed to the end of the array.

The test snippet that throws error:
```javascript
var holder = {"a": "1"};
fluid.pushArray(holder, "a", 3); // holder.array now holds [3]

console.log(holder);
```

Is this the use case where this function should not be used? Or the function should actually turn the leaf value into an array, that means the above example should output `{"a": ["1", 3]}`?

src/documents/InfusionEventSystem.md
```diff
@@ -210,18 +212,17 @@ Once an event firer is constructed, it can be called with the following methods
...
-    <td><code>listener: Function,String,
-listener specification
-</code></td>
+    <td><code>(listener {Function|String}[, namespace {String}, <br/> priority {String|Number}])</code></td>
     <td>
       Registers the supplied listener with this firer. The listener represents a function of a particular signature which is determined between the firer and listener of an event.
       The <code>namespace</code> parameter is an optional <code>String</code> which defines a key representing a particular <em>function</em> of the listener.
-      At most one listener may be registered with a firer with a particular namespace. This is a similar system to that operated by the JQuery namespaced events system.
+      At most one listener may be actively registered with a firer with a particular namespace. If a further listener B is registered with the same namespace as a previous one A, A will be
+      pushed onto a hidden stack, from the top of which it will be recovered by a call to <code>removeListener</code> removing B. Priorities are discussed in more detail on their own page [Priorities](Priorities.md).
```
cindyli: Because this paragraph is placed in html tags, replace the use of [Priorities](Priorities.md) to pure html using `<a>`.

src/documents/InfusionEventSystem.md
```diff
@@ -210,18 +212,17 @@ Once an event firer is constructed, it can be called with the following methods
...
     </td>
   </tr>
   <tr>
     <td><code>removeListener</code></td>
-    <td><code>listener: String/Function</code></td>
+    <td><code>(listener {String|Function})</code></td>
     <td>
       Supplies either the same listener object which was previously supplied to <code>addListener</code>, or else the String representing its namespace key. The designated listener will be removed from the list of registered listeners for this fierer.
```
cindyli: Typo on "fierer"? should be "firer"? same typo of the last word at line 713.

src/documents/InfusionEventSystem.md
```diff
@@ -262,14 +263,14 @@ var myFirer = fluid.makeEventFirer(options);
     </tr>
   </thead>
   <tr>
-    <td><code>preventable</code> (optional) DEPRECATED</td>
+    <td><code>preventable</code> [optional] DEPRECATED</td>
```
cindyli: Line 251 says: **The signature of `fluid.event.makeEventFirer` is not stable and will be revised in the 2.0 release of Infusion.** If the revised part is the deprecated `preventable` argument, shall this line be removed?

Also, the structure of `options` argument supplied to `fluid.makeEventFirer()` function is not clear. The doc probably wants to indicate that it's an object including members of elements listed in this table.

src/documents/ModelRelay.md
```diff
@@ -4,43 +4,42 @@ layout: default
...
+Every Infusion component descended from the grade `fluid.modelComponent` (a model-bearing component) supports a configuration area named `modelRelay` in which these rules can be defined, as well as allowing a short-form _implicit syntax_ in the `model` area which can be used where the model values to be synchronised are identical at each end of the relationship.
 
 ### Two styles for configuring model relay ###
 
-There are two primary styles of configuration for setting up model relay &#8212; firstly, using the _implicit syntax_ which just consists of [IoC References](IoCReferences.md) from the model configuration for one model-bearing component to another &#8212;
-that is, in the component's configuration under the top-level model entry. Secondly, the _explicit syntax_ involves an entry in the component's top-level `modelRelay` entry expressing a more complex rule, 
+There are two styles of configuration for setting up model relay &#8212; firstly, using the _implicit syntax_ which just consists of [IoC References](IoCReferences.md) from the model configuration for one model-bearing component to another &#8212;
```
cindyli: It would be helpful to add links on "implicit syntax" and "explicit syntax" to point to where they are explained down below in this doc.

src/documents/ModelRelay.md
```diff
@@ -4,43 +4,42 @@ layout: default
 category: Infusion
 ---
 
-The Infusion Model Relay system is a powerful scheme for supplying declarative configuration that specifies rules for keeping multiple pieces of model state around the component tree automatically up to date with each other's changes. 
+Using the Infusion Model Relay system, you can supply declarative configuration which specifies rules for keeping multiple pieces of model state around the component tree automatically up to date with each other's changes. 
 
-We sometimes use the term _**model skeleton**_ to refer to a connected set of models around the component tree which are related together by relay rules. This appeals to the idea that these models form an inner core which
+We sometimes use the term _**model skeleton**_ to refer to a connected set of models around the component tree which are related together by relay rules. This illustrates the idea that these models form an inner core which
```
cindyli: In the sentence "This illustrates the idea that these models form an inner core which arrive...", shall "arrive" be "arrives"?

src/documents/ModelRelay.md
```diff
@@ -134,20 +145,48 @@ This is done using the `modelRelay` section of a component's top-level options.
         </tr>
```
cindyli: Since I cannot add comment on line 141 which isn't modified in this pull request, leave the comment for it here on this closest line. 

I kind of remember you mentioned `modelRelay` block does not support nested transforms, which my understanding was that `transform` option is not supported. Am I wrong?

src/documents/InfusionEventSystem.md
```diff
@@ -162,26 +164,26 @@ In this case, you can use the long form record, where the listener record takes
   </thead>
   <tr>
     <td><code>listener</code>, <code>func/funcName</code></td>
-    <td><code>String</code></td>
+    <td><code>{String}</code></td>
     <td>This holds the actual designation of the function which is to be the listener, which was the string that consisted of the entire "short form" for the listener definition above. This, as before, holds either
     the name of a global function or an IoC reference to a function. The `func`/`funcName` form is supported for consistency with the syntax for [Invokers](Invokers.md) and the name `listener` will be deprecated for the Infusion 2.0 release.
```
cindyli: There are quite a number of occurences in this document that markdown symbols are used inside of html. They should be replaced with `<code>` or `<a>` etc.

src/documents/ModelRelay.md
```diff
@@ -134,20 +145,48 @@ This is done using the `modelRelay` section of a component's top-level options.
...
+            <td><code>{excludeSource: "init"}</code></td>
+        </tr>
+        <tr>
+            <td><code>namespace</code> (optional)</td>
+            <td><code>String</code></td>
+            <td>Identifies this model relay rule uniquely within its parent component. This namespace may be used to override the relay rule from a derived <a href="ComponentGrades.md">grade</a>, or to control the priority of
+            one relay rule with respect to another by using the <code>priority</code> field. 
+            Compare with the similar uses of namespaces within Infusion <a href="InfusionEventSystem.md#full-listener-record-form">events</a>, <a href="InfusionEventSystem.md#full-listener-record-form">model listeners</a>, etc. As with `modelListener`, this
+            value will be taken from the key of the <code>modelRelayBlock</code> if it is in the long form.</td>
```
cindyli: Can you elaborate this line: `As with modelListener, this value will be taken from the key of the modelRelayBlock if it is in the long form.`?

BTW, replace the markdown syntax around **`modelListener`** with `<code>` tag.

src/documents/ModelRelay.md
```diff
@@ -187,22 +226,38 @@ the updates will propagate only in one direction. This can still be highly usefu
...
-invertible, the `backward` option is ignored, and the system will behave as if it read `backward: "never"`. The control which can be exerted is over whether the relay is operated either during the
-<a href="#the-initial-transaction">initial transaction</a>, the "mature life" of the component after construction, or neither, or both. The default value for these options is `"always"` which
-implies that the system will always operate the relay rule if it can. 
+Each explicit relay rule can accept options `forward` and `backward` which allows the configurer to 
+control the occasions on which the relay is operated in those directions &#8212; that is, `forward` 
+representing the direction from `source` to `target`, and `backward` representing the direction 
+from `target` to `source`. If the transform is not invertible, the `backward` option is ignored, 
+and the system will behave as if it read `backward: {excludeSource: "*"}`.
+ 
+Some use cases for controlling propagation relative to the `init` source are:
```
cindyli: Would be nice to explain these use cases with code examples.

cindyli: I see, at the end of this section, there's a link pointing to the model listener examples that explain the same concept.

src/documents/CoreAPI.md
amb26: Not a typo, and the behaviour on foreign context objects is a crucial part of the contract. For example a simple check "x instanceof Object" will reject as plain an object which has come from another context, since it is descended from a different instance of the Object prototype. I removed the confusing word "even" - let me know if you think it could be improved further.

src/documents/CoreAPI.md
amb26: I've clarified that this is simply a wrapper for the builtin operator +.

src/documents/CoreAPI.md
amb26: I've added a sentence to clarify that the member must either be undefined or hold an array.

src/documents/InfusionEventSystem.md
amb26: This link works fine in the docpad generated markup. There are several oddities with the markup processing with github which mean that we cannot always depend on it - for example, it uses a different algorithm for generating anchor ids.

src/documents/InfusionEventSystem.md
amb26: I've added a table heading, and removed the section about the 2.0 releases

src/documents/ModelRelay.md
amb26: Arrive is in theory correct, since it is the models which arrive - especially since the next verb comes with "before they start". However the sentence is unclear and I've tried to improve it a bit. Still not great. If only we had a _technical writer_.

src/documents/ModelRelay.md
amb26: It is to some extend supported since it is tested here: https://github.com/fluid-project/infusion/blob/master/tests/framework-tests/core/js/DataBindingTests.js#L1154 - I don't know of a case where it is being used in the wild and we do somewhat discourage it - see IRC discussion at https://botbot.me/freenode/fluid-work/2016-04-07/?msg=63718987&page=2

We expect most cases that might currently require nested transforms to be taken care of by upcoming work on the REDUCED EXPRESSION PARSER - https://issues.fluidproject.org/browse/FLUID-5894 - the facility for nested transforms in relays is left around in "hopeful exploration" to see if anyone can claim a use for it.

src/documents/InfusionEventSystem.md
amb26: Noted above - these symbols work fine in the final docpad output.

src/documents/ModelRelay.md
amb26: Good idea - I've added a (working!) sample. wrt the model listener examples - this has only become "the same concept" as a result of this rewrite - it's also not precisely the same since a source value of "local" is meaningless for a model relay rule.

amb26:
Thanks - ready for further review

cindyli:
Merged at 9fed41658edc8bd560dce4f39a1611de75608d01

PR #85, opened 2016-03-22
FLUID-5880: Updating the Localization Docs

jobara:
Updated the Localization in the Preferences Framework docs based on changes from FLUID-5597 and FLUID-5599 regarding the message bundles.


src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -72,9 +72,9 @@ The message bundle is attached to each panel component as the `parentBundle` opt
 Message bundles can be specified in one of two ways:
 
 1. through the [auxiliary schema](AuxiliarySchemaForPreferencesFramework.md) (if schemas are being used), or
-1. directly to the `messageLoader` (if grades are being used).
+2. directly to the `messageLoader` (if grades are being used).
 
-The Preferences Framework will load and combine all of the Message Bundles into a single Message Bundle which is bound to the `prefsEditorLoader` component at the `msgResolver` property (as described above).
+The Preferences Framework will load all of the Message Bundles. When using an auxiliary schema the bundles are automatically distributed to the panels. If an auxiliary schema is not used, the `messageBase` options of each panel will need to be set manually using an IoC reference; `{messageLoader}.resources.<messageBundle>.resourceText`.
 
 Any panel that has the grade `fluid.prefs.defaultPanel` will have access to the combined Message Bundle at its `parentBundle` option (as described above). When using the auxiliary schema, all panels are assigned the grade `fluid.prefs.defaultPanel` by the Framework.
```
cindyli: We probably should remove the saying of ```parentBundle```? With our first, and old, localization implementation, users were told to IoC referencing to ```parentBundle```for retrieving messages. This is now all transparent to users. Continuing having it here only confuses readers, i think.

cindyli: Down below in "Example Auxiliary Schema" section, can we add comments to parts that are localization related, such as the message bundle files specified at the prefsEditor level and panel levels.

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -72,9 +72,9 @@ The message bundle is attached to each panel component as the `parentBundle` opt
 Message bundles can be specified in one of two ways:
 
 1. through the [auxiliary schema](AuxiliarySchemaForPreferencesFramework.md) (if schemas are being used), or
-1. directly to the `messageLoader` (if grades are being used).
+2. directly to the `messageLoader` (if grades are being used).
```
cindyli: I'm thinking if we should remove grade version of preferences framework from this doc. I looked thru our other docs for preferences framework, the grade version is not mentioned elsewhere at all. Sort of remember we decided to promote schema version since it's more integrator friendly.

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -130,9 +130,9 @@ Any panel that has the grade `fluid.prefs.defaultPanel` will have access to the
 ```javascript
 fluid.defaults("my.messageLoader", {
```
cindyli: Shall we remove the example of creating message loader too? As in the above comment, if the grade version of preferences framework was no longer recommend to integrators, integrators would not need to create their own message loaders any more.

cindyli:
At line 206-208 that specifies:
```
     stringArrayIndex: {
         theme: ["contrast-default", "contrast-bw", "contrast-wb", "contrast-by", "contrast-yb", "contrast-lgdg"]
     }
```

The array values in the ```theme``` option maps to the message key defined in the .json file. This relationship should be documented.

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -232,3 +232,29 @@ The strings can also be accessed directly, outside of the context of IoC referen
 ```javascript
 that.msgLookup.lookup(value); // where value is either the string name or the key in the stringArrayIndex to lookup.
```
cindyli: Is this still valid?

src/documents/LocalizationInThePreferencesFramework.md
jobara: looks like `fluid.prefs.defaultPanel` doesn't exist anymore at all, so i'll just remove that whole section.

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -232,3 +232,29 @@ The strings can also be accessed directly, outside of the context of IoC referen
...
+## Specifying a Localization
+
+The messageLoader takes `defaultLocale` and `locale` options for specifying which localized message bundle to fetch. The `locale` option specifies the localization desired. By default it is sourced from the prefsEditorLoader's settings object, `"{prefsEditorLoader}.settings.locale"`. The `defaultLocale` provides a fallback to use if the desired localization cannot be located. By default it is sourced from the prefsEditorLoader's `defaultLocale` option.
+
+### Fallback Rules
+
+If a requested localization cannot be found, the messageLoader will attempt to find another message bundle to satisfy the request. Language codes are expected in a form similar to BCP 47 tags but with a "\_" instead of a "-" separating the language code from the country code.
+
+Assuming the Canadian French message bundle is requested:
+1. look for a message bundle corresponding to the language code specified by the `locale` option (e.g. "fr_CA")
```
cindyli: The formatting is incorrect when viewing in github. All lines with numbered points are joined together as one big block. See https://github.com/jobara/infusion-docs/blob/FLUID-5880/src/documents/LocalizationInThePreferencesFramework.md

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -232,3 +232,29 @@ The strings can also be accessed directly, outside of the context of IoC referen
...
+2. look for a message bundle with the same language as the language code specified by the `locale` option (e.g. "fr")
+3. look for a message bundle corresponding to the language code specified by the `defaultLocale` option (e.g. "en_US")
+4. look for a message bundle with the same language as the language code specified by the `defaultLocale` option (e.g. "en")
+5. look for a message bundle with the exact URL as specified through the auxiliary schema or directly to the messageLoader (e.g. "message/prefsEditor.json")
```
cindyli: Remove "or directly to the messageLoader" if we decide not to mention the grade version of prefs framework.

src/documents/LocalizationInThePreferencesFramework.md
jobara: okay

cindyli:
Seems you accidentally committed "PrefsFrameworkLocalization.graffle".

src/documents/LocalizationInThePreferencesFramework.md
jobara: Looks like it still is. https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/js/Panels.js#L18-L53

cindyli: But it seems we don't quite use it in the code. Instead, the IoC references like ```{that}.msgLookup.slidingPanelShowText``` are used.

jobara:
PrefsFrameworkLocalization.graffle was already there. I updated it though with the changes for the diagram.

src/documents/LocalizationInThePreferencesFramework.md
cindyli: After thinking it more, we probably should keep it since it does provide another alternative to retrieve message strings. But make it clear that the prerequisites for using this function is to have ```fluid.prefs.msgLookup``` as a grade component. This statement is useful when integrators create their own sub-components of ```prefsEditorLoader```.

cindyli:
Thanks for clarifying the existence of PrefsFrameworkLocalization.graffle. It's great to keep the source file along with the converted images. Never realized it's been there. Thanks.

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -34,49 +36,40 @@ Message bundles cannot contain arrays. Instead a namespace should be used to gro
...
 
-- parses and assembles JSON strings loaded by the MessageLoader,
-- runs the assembled JSON through the [message resolver](http://wiki.fluidproject.org/display/docs/fluid.messageResolver) to create the lookup function, and
-- attaches the message resolver bundle as a member, accessible through `that.msgResolver`.
+### MessageLoader
 
-To access the message bundle from other components on the IoC tree, use `{prefsEditorLoader}.msgResolver`.
+The messageLoader is the fluid.prefs.resourceLoader responsible for retrieving all of the message bundles added to the Prefs Editor (See [Adding Message Bundles](#adding-message-bundles) below). Message bundles are automatically distributed to the Prefs Editor and Panels based on the configuration supplied in the [auxiliary schema](AuxiliarySchemaForPreferencesFramework.md). However, to access the message bundle from other components on the IoC tree, the IoC reference `{messageLoader}.resources.<messageBundle>.resourceText` can be used. Additionally the `defaultLocale` and `locale` options are used to specify which localization to fetch for each bundle. (See: [Specifying a localization](#specifying-a-localization) below)
```
cindyli: Shall we say: The messageLoader is an instance of fluid.prefs.resourceLoader ...

cindyli:
Thanks for the update, @jobara.

As discussed, in the section of "Preferences Editor Component Hierarchy", the explanations for PrefsEditorLoader and prefsEditor are no longer necessary. They were there because with the old strategy, we had different ways for retrieving message strings at different hierarchy level, which is no longer valid. Instead, the doc for ```fluid.prefs.msgLookup``` and how to use it can be added here, along with ```messageLoader``` that is already here. These components are essential for understanding the current implementation of how to localize the prefs framework. 

Moreover, the steps we talked about of how to use ```msgLookup``` on integrators' own components:
1. add ```fluid.prefs.msgLookup``` as a grade component;
2. for non-renderer components, add a ```msgResolver``` as a subcomponent with a proper ```messageBase``` option;
3. for renderer components that automatically receive a ```msgResolver```, pass the correct```resourceText``` to the ```messageBase``` option of the ```msgResolver``` component.

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -32,51 +34,70 @@ Message bundles cannot contain arrays. Instead a namespace should be used to gro
...
+The `fluid.prefs.msgLookup` grade is required for any component through which a message lookup is performed. The msgLookup grade wires up the `msgLookup` IoC reference and the `lookup` method for retrieving messages from the Message Bundle (See: [Using Message Bundles](#using-message-bundles)). The msgLookup's message resolution is mediated by a `fluid.messageResolver` subcomponent and it is to this subcomponent that the `messageBase` option, containing the fetched Message Bundle, is set.
 
-- parses and assembles JSON strings loaded by the MessageLoader,
-- runs the assembled JSON through the [message resolver](http://wiki.fluidproject.org/display/docs/fluid.messageResolver) to create the lookup function, and
-- attaches the message resolver bundle as a member, accessible through `that.msgResolver`.
+For a `fluid.prefs.panel` component all of the `fluid.prefs.msgLookup` configuration is prewired. For other components to make use of this functionality the following configuration is needed:
```
cindyli: A missing comma after "For a `fluid.prefs.panel` component".

cindyli: Another missing comma after "For other components to make use of this functionality".

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -32,51 +34,70 @@ Message bundles cannot contain arrays. Instead a namespace should be used to gro
...
+- add `fluid.prefs.msgLookup` as a [parent grade](ComponentGrades.md#specifying-parent-grades)
+- add an instance of `fluid.messageResolver` as a subcomponent with the member name `msgResolver`
+- link the messageBase to the fetched Message Bundle.
+    - e.g. using an IoC reference to the [messageLoader](#messageloader): `{messageLoader}.resources.<messageBundle>.resourceText`
+- for an instance of `fluid.rendererComponent` it is also necessary to use the messageResolver's `resolve` method as the `messageLocator`
 
-### PrefsEditor
-
-PrefsEditor is the host component that holds all the actual panel (or adjuster) components as subcomponents. By default, the message bundle is *not* passed down to PrefsEditor. If your PrefsEditor component will need direct access to the message bundle, provide it at the instantiation of any PrefsEditor instance, as shown in the following example:
+#### Adding MsgLookup to a Component
```
cindyli: Shall it say "Adding MsgLookup to a non-renderer Component" so users don't misunderstand it's only for ```fluid.component```?

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -32,51 +34,70 @@ Message bundles cannot contain arrays. Instead a namespace should be used to gro
...
-        msgResolver: "{prefsEditorLoader}.msgResolver"
+fluid.defaults("my.component", {
+    gradeNames: ["fluid.prefs.msgLookup", "fluid.component"],
+    components: {
+        msgResolver: {
+            type: "fluid.messageResolver"
+        }
+    },
+    // should contain the strings loaded from a Message Bundle
+    // e.g. "{messageLoader}.resources.messageBundleName.resourceText"
```
cindyli: Probably use a complete line as an example: ```messageBase: "{messageLoader}.resources.messageBundleName.resourceText"```.

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -32,51 +34,70 @@ Message bundles cannot contain arrays. Instead a namespace should be used to gro
...
-If the message bundle is provided to PrefsEditor this way, access it within the PrefsEditor component using `{that}.options.msgResolver`.
+#### Adding MsgLookup to a Renderer Component
 
-### Panels
-
-The message bundle is attached to each panel component as the `parentBundle` option. To access it from within a panel, use `{that}.options.parentBundle`.
+```javascript
+fluid.defaults("my.renderer.component", {
+    gradeNames: ["fluid.prefs.msgLookup", "fluid.rendererComponent"],
+    components: {
```
cindyli: Is it still needed to manually add ```msgResolver``` sub-component if using ```fluid.rendererComponent```?

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -32,51 +34,70 @@ Message bundles cannot contain arrays. Instead a namespace should be used to gro
...
+        msgResolver: {
+            type: "fluid.messageResolver"
+        }
+    },
+    rendererOptions: {
+        messageLocator: "{msgResolver}.resolve"
+    },
+    // should contain the strings loaded from a Message Bundle
+    // e.g. "{messageLoader}.resources.messageBundleName.resourceText"
```
cindyli: Same as the comment above to use the full line to clarify.

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -196,7 +198,7 @@ fluid.defaults("fluid.slidingPanel", {
 There are other, more complex cases where an array of strings is required (for example, for a set of radio buttons or a drop-down). In these cases, a `stringArrayIndex` in the components options needs to be specified. This defines both
 
 1. which strings to include and
```
cindyli: A missing comma before "and"?

src/documents/LocalizationInThePreferencesFramework.md
jobara: Yes it seems so. https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/js/Panels.js#L59-L82

src/documents/LocalizationInThePreferencesFramework.md
jobara: I understand what you mean, but then the order of the two sections seems strange. That is the render component section should probably come first, but this seems strange as it's the odd case. What do you think? Any other ideas?

src/documents/LocalizationInThePreferencesFramework.md
```diff
@@ -232,3 +248,30 @@ The strings can also be accessed directly, outside of the context of IoC referen
...
 that.msgLookup.lookup(value); // where value is either the string name or the key in the stringArrayIndex to lookup.
 ```
+
+## Specifying a Localization
+
+The messageLoader takes `defaultLocale` and `locale` options for specifying which localized Message Bundle to fetch. The `locale` option specifies the localization desired. By default it is sourced from the prefsEditorLoader's settings object, `"{prefsEditorLoader}.settings.locale"`. The `defaultLocale` provides a fallback to use if the desired localization cannot be located. By default it is sourced from the prefsEditorLoader's `defaultLocale` option.
+
+### Fallback Rules
+
+If a requested localization cannot be found, the messageLoader will attempt to find another Message Bundle to satisfy the request. Language codes are expected in a form similar to BCP 47 tags but with a "\_" instead of a "-" separating the language code from the country code.
```
cindyli: Can you add a link for "BCP 47 tags"? Thanks.

src/documents/LocalizationInThePreferencesFramework.md
cindyli: I don't have a good idea for this. Probably still use "Adding MsgLookup to a component" but add a note in the section that indicates renderer components need special handling and this approach is for other component types?

PR #86, opened 2016-04-14
FLUID-4030: All transform functions are now documented

kaspermarkus:


src/documents/ComponentConfigurationOptions.md
```diff
@@ -382,8 +382,9 @@ See also: [Component Grades](ComponentGrades.md)
         source: "volumeAsPercent",
         target: "volumeAsFraction",
         singleTransform: {
-            type: "fluid.linearScale",
-            factor: 0.01
+            type: "fluid.transforms.linearScale",
+            factor: 0.01,
+            inputPath: ""
```
amb26: This can't be right - we don't use the "xxxPath" form in relay documents. This can be removed again

src/documents/ModelRelay.md
```diff
@@ -120,8 +120,9 @@ This is done using the `modelRelay` section of a component's top-level options.
             <td>A short form which can be used where the transformation consists of just a single Model Transformation transform rule. Use either this or <code>transform</code></td>
             <td><code>
 {
-    type: "fluid.linearScale",
-    factor: 100
+    type: "fluid.transforms.linearScale",
+    factor: 100,
+    inputPath: ""
```
amb26: Remove inputPath

src/documents/ModelRelay.md
```diff
@@ -164,8 +165,9 @@ fluid.defaults("examples.explicitModelRelay", {
                     source: "{examples.explicitModelRelay}.model.volume",
                     target: "volume",
                     singleTransform: {
-                        type: "fluid.linearScale",
-                        factor: 0.01
+                        type: "fluid.transforms.linearScale",
+                        factor: 0.01,
+                        inputPath: ""
```
amb26: remove inputPath

amb26:
Commenting with line numbers on ModelTransformationAPI.md since the file/diff is too big to appear in review interface:

8: We don't require that the markup renders perfectly in github's preview, but we should try to remove gratuitous incompatibilities. Change the *** bolding to remove the spaces between the asterisks and the bolded text
12: Be clear that we don't support completely equivalent features in the modelRelay context as in the transform context. In particular, we don't support xxxPath in modelRelay, and we steer towards the use of SingleTransform rather than full transforms.
20-41: Please convert these into links to the respective sections. A slightly quick way to do this is to copy the "link icon" target from the fully rendered docpad view
58 and following: Remove the json type from the quoted text since it causes a corrupt-looking view in github and doesn't do anything useful in the fully rendered markup (same for all such blocks)
77: Stray link left back to the GPII wiki - all docs should be here now
82: Clarify: reserved words, considered as JSON keys
78: This isn't really quite right. Since we implemented https://issues.fluidproject.org/browse/FLUID-5867 , the syntax of nested transform material is not the same as that of top-level material. You can't get back into the outputPath: inputPath world once having left it. This implies we need three general "grammar states" - TopLevelTransform = PathTransform|FullTransform, and SingleTransform
91: "types" is probably a clearer standard term than "classes"
100: This behaviour is insane and i) we shouldn't advertise it, ii) we should make it an error to specify both inputPath and input in the same transform iii) we should check that we haven't blundered and made any piece of impl rely on this behaviour
114: grade names in monospace
118: monospace rather than quotes
132: It doesn't, it seems to instead have a member named "randomSeed"
138: or inputPath, rather than and
153: Throughout this section, quote config keys using monospace rather than " quotes
159: So will.... be - unclear sentence
470: different -> difference
750: Be a bit clearer - the framework can generate the inverse of a document where it doesn't use nested transforms, and to the extent that individual transforms are invertible.
752: inversed -> inverted
754: -> "In practice, perfect inversion of a rule is not always possible".
758: Just put the bolding around the title of the item rather than the formula since it stuffs up in github
762: exit -> exist
762: It may be that `G` is defined, but simply hasn't been registered in the framework.
744: No, it is not! This is very confusing. fluid.transforms.identity is the identity function, that returns its input unchanged. literalValue is a constant function, that takes the same value on any input.
857: Not right, and confusing - fluid.transforms.value is perfectly invertible, by definition. This text "the input default values are always ignored" which appears for many transforms is confusing and should be removed. The "input" values *are* the input values, that need to be reproduced - they are never ignored!
996: "lacky" is not really a word. Perhaps, "lossy"? Also, "if the input is not a valid number" is confusing, given you are talking about the inverse. It would be clearer to start talking using standard terms "domain" and "range" https://en.wikipedia.org/wiki/Domain_of_a_function and say things like "is invertible when its domain is restricted to numbers"
1195: Not true any more, see notes on FLUID-5867
1326: "will do a mapping based on the input value" doesn't really say anything. Perhaps, "map a defined collection of input values to corresponding output values by applying a matching algorithm".
1356: Consistency of quoting throughout this section - use monospace rather than "
1357: Stray colon
1597: Remove all the colons following the operators, which are very confusing
1599: This should be escaped as an HTML entity
1621: Missing newline after triple quotes
1699: Monospace types like `multiInputTransformFunction`. Extra newline before "Syntax"
1705: Can't the condition be a path as well? I think it's clearer not to repeat all the "variable or path" business since it should be 
1853: This is a bit messy - we should just have a standard format for multiInputTransformFunction where we list what the inputs are, and put defaults, type info, etc. in the transform docs. Also, transform docs should take the form of a table with 3 columns: name, type, description as we use elsewhere in the framework rather than the ad-hoc "JSON with angle brackets" form
1871: variables
1873: Something's gone wrong with all these tables in that the output appears outside it. Missing td opening tag I think
2000: "Output" column of these tables has stuffed up as before
2140: Convert to 3-column table rather than "JSONish" as above
2526: As above, better to say "fully invertible over domain X" rather than "partly invertible"
2827: Block not quoted (see following ones too)
3207: Check all "output" columns in the document

kaspermarkus:
Thanks for all the comments on this pull, Antranig.. Addressed most of the stuff now, but have some questions:

> (on inputPath vs input)  100: This behaviour is insane and i) we shouldn't advertise it, ii) we should make it an error to specify both inputPath and input in the same transform iii) we should check that we haven't blundered and made any piece of impl rely on this behaviour

Why is this behavior insane? Isn't is useful to be able to provide a default fallback value? I dont think any behavior relies on this functionality, but the handling of both supporting input and inputPath (even simoultanously) is boiled into the framework.



> (on 3 types in the inversion section) 762: It may be that G is defined, but simply hasn't been registered in the framework.

I'm not sure I understand this comment... Are you saying that, a G (inverse) might exist, but it just hasn't been implemented?



> 857: Not right, and confusing - fluid.transforms.value is perfectly invertible, by definition. This text "the input default values are always ignored" which appears for many transforms is confusing and should be removed. The "input" values are the input values, that need to be reproduced - they are never ignored!

I think this might be one of those cases where I'm totally confusing myself. But I'm not sure what a "input" would map to in the inverse. Obviously, inputPath would map to outputPath. Also, re: inversions, things become a bit screwy when both input and inputPath are defined (see FLUID-5337)


> (Re: linear scale docs) 1853: This is a bit messy - we should just have a standard format for multiInputTransformFunction where we list what the inputs are, and put defaults, type info, etc. in the transform docs. Also, transform docs should take the form of a table with 3 columns: name, type, description as we use elsewhere in the framework rather than the ad-hoc "JSON with angle brackets" form

Not sure I understand -- do you want us to have this in the documentation, ie: "a table with 3 columns: name, type, description". All the transforms documentation has the syntax section with JSON - should these be removed also (and replaced with tables)

javihernandez:
I've spotted a couple of things in the valueMapper documentation.
* 1484: outputPath must be "myPath2" if we want the output to be like that
* 1514: defaultInputPath must be "whichAnimal" according to the source provided

src/documents/ModelRelay.md
```diff
@@ -120,8 +120,8 @@ This is done using the `modelRelay` section of a component's top-level options.
             <td>A short form which can be used where the transformation consists of just a single Model Transformation transform rule. Use either this or <code>transform</code></td>
             <td><code>
 {
-    type: "fluid.linearScale",
-    factor: 100
+    type: "fluid.transforms.linearScale",
+    factor: 100,
```
cindyli: Remove the comma.

kaspermarkus: fixed (github doesn't seem to update this pull request, though)

cindyli:
12: missing period at the end of the paragraph.
100: missing colon at the end of the paragraph.
127: remove the extra period at the end of the paragraph.
147: change `standardInputTransformFunction` `andmultiInputTransformFunction` -> `standardInputTransformFunction` and `multiInputTransformFunction` (meaning, moving the word "and" out of the code block)
149: missing colon after "in the following way". Also, missing closing bracket for "(as for the standardInputTransformFunction". And, add a link on "the multiInputTransformFunction section" that directs readers to that section.
170: remove empty spaces before and after the word "Example" for the markdown syntax "***" to be rendered properly.
217, column 45: quote the word "Magnification" in a code block as it's a key specified in the example.
296, column 207: The code block for `outputPath: dataType` is messed up
342: "in turns" -> "in turn".
477, column 136: quote `undefined` in a code block.
479: different -> difference, the previous sections -> the previous section
567: the output will way -> the output will vary
571: At "(see next section)", Add a link for "next section" to point to where the next section is
761: missing period at the end of the paragraph.
768, 770, 772: 2 stars (`**`) at the beginning of the paragraph has no matching closing 2 stars. 
768: Missing period at the end of the paragraph.
771: (or hasn't been defined in the fram -> (or hasn't been defined in the framework)
772: This sentence is not clear: `Non invertible are just that: not invertible`
850, 851: fix indentations and add a closing curly bracket.
864: This transforms -> This transform
893-897 & 923-927: The output at these 2 spots should be:
```
{
    "myfavorite": "balloon"
}
```
1154-1156: The output should be:
```
{
    "howLong": 1
}
```
1204: Will return -> Returns (to be consistent with other descriptions)
1282: This transformation will delete -> Deletes (same reason as above)
1299: remove this line because there's only one example for deleting path
1340, 1363: remove the unnecessary `<code>` and `</code>`
1383: Capitalize the first letter of the sentence. Missing period at the end of the sentence.
1384: Missing period at the end of the sentence.
1387: Use code quotation for the word "defaultOutputPath" instead of single quotes
1389: Use code quotation for the word "outputPath" instead of single quotes
1390: Missing period at the end of the sentence.
1394: Capitalize the first letter of the sentence. Use code quotation for the word "defaultInputPath" instead of single quotes. 
1396: as output value -> as an output value. Also, from what's described in the line, `outputUndefinedValue` has higher priority than `outputValue` and `defaultOutputValue`. What if `outputPath` is specified? With these many keys for the output, it's probably nice to explain their order of priority at parsing.
1398: Capitalize the first letter of the sentence to be consistent.
1406: ValueMapper support -> ValueMapper supports. are be interpreted -> are interpreted. This paragraph also mentions `the longhand syntax` and it seems example 3 at line 1500 and example 4 at line 1541 use this kind of syntax. Can the title of these examples be adjusted by having "the longhand syntax" in to be more specific and clear?
1599: it might be good enough to explain `(variables: left, right)` in the description below. A bit confusing to have them at there.
1636: The closing code quotation needs to be in a new line. Otherwise the following format is messed up as what's happening now.
1709: As above, I think it's less confusing to have "(variables: `true`, `false`, `condition`)" here.
1864: As above, thinking nicer to explain "variables: factor - default=1, offset - default=0" in the description instead of having it here.
2296: The last sentence in this paragraph is incomplete.
2298: Add a link for "deindexIntoArrayByKey" to point to that section.
2308: The closing code quotation needs to be in a new line. 
2314: as they key -> as the key
2397: creates -> create
2416: Missing period at the end of the sentence.
2417: wrap words "deindexIntoArrayByKey" and "*Path" into code quotations.
2420: The last sentence in this paragraph is incomplete.
2422: Add a link for "indexArrayByKey".
2432: The closing code quotation needs to be in a new line. 
2438: This sentence is unclear: This means that each of the keys within the object given as input, will be added as a value to the corresponding object with `product` as key.
2522: It is found -> If it is found. Also, add a comma after "If it is not found".
2530: Add a link for "dereference".
2704: Add a link for "indexOf".
2750: Does this example work for you? The transformation result I received is an empty object `{}`.
2823, 2827: The closing code quotation needs to be in a new line. 
2994: Add a link for "setMembershipToArray".
3102: into `setMembershipToArray` -> into `arrayToSetMembership`. Also, add a link for "setMembershipToArray".
3100: is based on -> are based on

kaspermarkus:
Thanks so much @cindyli for your comments - I've addressed everything with the exception of:

> 1396: (...) Also, from what's described in the line, outputUndefinedValue has higher priority than outputValue and defaultOutputValue. What if outputPath is specified? With these many keys for the output, it's probably nice to explain their order of priority at parsing.

Which I'm not sure I understand properly.. Are you suggesting that I put in a seperate place in the valuemapper description, an explanation of the priority of those keys?

cindyli:
Yes, @kaspermarkus, that's what I meant. 

I also think it might be worthwhile to have a separate section for all reserved keywords used by the model transformation. I understand these words could vary depending on which transform being used etc. But as a developer, I personally would avoid using those keywords completely instead of checking back on each transform for what can be used and what can not. Of course, this is just a personal opinion. Feel free to use other methods.

kaspermarkus:
@cindyli thanks for the input! I've updated according to your suggestions which I think are both good

kaspermarkus:
Fixed a formatting issue

@cindyli Dont know how relevant this pull request is for the infusion 2.0 release, but if it should be in by monday, if you could let me know any remaining issues by tomorrow morning my time, that'd be great - I dont think I'll have time to do any changes over the weekend

cindyli:
Thanks for the fix. The table of reserved words is great. Here comes more but most are minor:

13, 16, etc: When using "(see below)", add a link to "below" to point readers to where the information can be found. This applies to other "(see below)"s in this doc.
21: broken link for "Transformation Functions"
24: broken link for "fluid.transforms.identity"
25: broken link for "fluid.transforms.literalValue"
41: broken link for "fluid.transforms.deindexIntoArrayByKey"
52: remove the closing bracket ")" after the word "below". There are one opening bracket but two closing brackets here.
55: The last sentence in this paragraph says: "This option is unsupported." Why is this option on the doc if it's not supported?
82: which accept -> which accepts
156: the path to output to -> the path to output
164: missing period at the end of this paragraph. Also, probably remove "(see above)" and directly add links for `standardInputTransformFunction` and `standardOutputTransformFunction`
168: These transformations allows -> These transformations allow
188: an variable -> a variable; code quote the word "standardInputTransformations" at column 246; has other inputs that needs -> has other inputs that need
209: as so\ -> as so; is interpreted according the -> is interpreted according to the; it's content -> its content
256: So the walking through -> So walking through; code quote the word "Magnification" used at "which in turn is relative to Magnification".
258: "This means that" is used in two connected sentences. Probably change the second one to "as a result".
262: To be consistent: the class -> the grade
528: One can use the output of another transformation as an input value -> One can use the output of a transformation as an input value of its parent transformation.
608: code quote the word "literalValue" in the sentence "the result of the literalValue transformation"
612: broken link for "ssee next section". also typo: ssee -> see
802: The description on the single input of this function is confusing. The input could be the transformation document or transformation rules? These are different concepts. Same question for the output.
809: matches the ones -> match the ones
820: Probably switch the order of descriptions for fluid.transforms.literalValue() and fluid.transforms.value() to match up the order in the table of content for "transformation functions" at the top of the doc
1424: Capitalize the first letter of this sentence
1752, 1907: remove the code quotes around grades specified in "type" sentence to be consistent with "type" blocks of other transform functions
1766: Shall the single quotes used for '*Path' be replaced with code quotes? Also applies to the entire doc where single quotes are used.
2335: an that -> and that

cindyli:
@kaspermarkus, when generating html and running docpad server locally, this pull request produces an error that doesn't occur in the master branch:

```
warning: A warning occured:

Error: Highlight.js: Syntax highlight definition does not exist for language: lang-{
Your probably want to add an alias for this language to one that we do support
More information here: https://github.com/docpad/docpad-plugin-highlightjs#language-aliases
  at DocPad.warn (/usr/local/lib/node_modules/docpad/out/lib/docpad.js:3541:50)
  at HighlightjsPlugin.highlightSource (/Users/cindyli/Development/infusion-docs/node_modules/docpad-plugin-highlightjs/out/highlightjs.plugin.js:92:24)
  at replace (/Users/cindyli/Development/infusion-docs/node_modules/docpad-plugin-highlightjs/out/highlightjs.plugin.js:141:33)
  at Task.<anonymous> (/Users/cindyli/Development/infusion-docs/node_modules/bal-util/out/lib/html.js:73:20)
  at ambi (/Users/cindyli/Development/infusion-docs/node_modules/ambi/source/index.js:84:3)
  at Domain.fireMethod (/Users/cindyli/Development/infusion-docs/node_modules/taskgroup/out/lib/taskgroup.js:397:23)
  at Domain.run (domain.js:221:14)
  at Task.fire (/Users/cindyli/Development/infusion-docs/node_modules/taskgroup/out/lib/taskgroup.js:435:27)
  at Immediate.<anonymous> (/Users/cindyli/Development/infusion-docs/node_modules/taskgroup/out/lib/taskgroup.js:452:26)
  at runCallback (timers.js:574:20)
  at tryOnImmediate (timers.js:554:5)
  at processImmediate [as _immediateCallback] (timers.js:533:5)
```

kaspermarkus:
ugh, sorry for missing that... Should be fixed now @cindyli

cindyli:
Thanks for fixing the issue with building local docpad server, @kaspermarkus.

If you access "Model Transformation" page thru a local docpad server, there are some broken links:

23: fluid.transforms.value
24: fluid.transforms.identity
25: fluid.transforms.literalValue
41: fluid.transforms.deindexIntoArrayByKey

kaspermarkus:
@cindyli not sure how I missed those - I was convinced I already did them.. Anyway, apologies for missing them. Updated pull

cindyli:
Merged at 186b4376d61c187f743e20f7045d128f03d7051c

PR #93, opened 2016-07-11
FLUID-5892: add documentation of context-awareness behaviour for prefs adjuster widgets

waharnum:
This adds some documentation of switching between native and jQuery adjuster widgets for the now-merged PR: https://github.com/fluid-project/infusion/pull/724

### DIFF
--- a/src/documents/Builder.md
+++ b/src/documents/Builder.md
@@ -103,7 +103,7 @@ var myBuilder = fluid.prefs.builder({
 ```
 It is not necessary to specify the primary schema; The builder will automatically find the preference specifications provided by the Framework and build a primary schema (see [Processing the Schemas](#processing-the-schemas) below for more information).
 
-#### Switching Between Native HTML and jQuery UI Widgets
+### Switching Between Native HTML and jQuery UI Widgets ###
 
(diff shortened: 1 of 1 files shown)


waharnum:
@jobara this is the documentation I mentioned

src/documents/Builder.md
```diff
@@ -103,6 +103,36 @@ var myBuilder = fluid.prefs.builder({
 ```
 It is not necessary to specify the primary schema; The builder will automatically find the preference specifications provided by the Framework and build a primary schema (see [Processing the Schemas](#processing-the-schemas) below for more information).
 
+#### Switching Between Native HTML and jQuery UI Widgets
```
jobara: Was this intended to be a subsection of "Using the Starter Preferences"? I feel like it could be at the same level, but can be convinced either way. Also could you add the #s to the end of the header too, just to make it easier to see in the file.

waharnum: Yes, it was intended as a subsection. I agree it could be at the same level - figuring out where it made the most sense to put it was a bit troublesome. Thinking on it now, I think it's probably best at the same level, since it applies to any instance where an adjuster widget with textfieldSlider is used, not just the starter preferences.

jobara:
Merged at f5af742ba3f4f4099d8027c391de14ab4e33074c

PR #94, opened 2016-07-26
FLUID 5935: Update the TextToSpeech component docs

waharnum:
This addresses two issues in our documentation about the TextToSpeech component:

* the current version of the component expects the speech options to be configured as part of the component model, while the current version of the documentation specifies configuring it as part of the top-level options block

* per the errata at https://dvcs.w3.org/hg/speech-api/raw-file/tip/speechapi-errata.html, `voiceURI` is no longer used by the `SpeechSynthesisUtterance` interface; instead, the `voice` attribute can be passed a `SpeechSynthesisVoice` object to configure the voice synth to be used, typically after retrieving it via the `getVoices()` method or similar.

Longer term we may want to refactor this component design - under the current implementation configuring the voice synthesizer requires storing the `SpeechSynthesisVoice` object on the model, and I'm not certain without further experimentation if this is a copy-safe operation (that is, does the `voice` attribute expect and validate for a `SpeechSynthesisVoice` object, or will any POJO with the correct attributes suffice). However, this PR brings the documentation in line with our own implementation and the current API.

### DIFF
--- a/src/documents/TextToSpeechAPI.md
+++ b/src/documents/TextToSpeechAPI.md
@@ -15,7 +15,7 @@ The Text To Speech component can be used in browsers that support [Web Speech Sy
 * MS Edge 14 (in Preview Release)
 * Firefox 48+ (using the `media.webspeech.synth.enabled` about:config option)
 
-<div class="infusion-docs-note"><strong>Note:</strong> Find the latest browser support data for Web Speech Synthesis API from
+<div class="infusion-docs-note"><strong>Note:</strong> find the latest browser support data for Web Speech Synthesis API from
 <a href="http://caniuse.com/#feat=web-speech">caniuse.com</a>.</div>
 
 ## Creator ##
@@ -230,7 +230,7 @@ fluid.queueSpeech("Hello world", false, {
 
 ## Supported Events ##
 
-<div class="infusion-docs-note"><strong>Note:</strong> If needed, please read the <a href="InfusionEventSystem.md">Infusion Event System</a> document for a full description of infusion events.</div>
+<div class="infusion-docs-note"><strong>Note:</strong> if needed, please read the <a href="InfusionEventSystem.md">Infusion Event System</a> document for a full description of infusion events.</div>
 
 The events fired by the Text To Speech component are described below.
 
@@ -379,11 +379,11 @@ The events fired by the Text To Speech component are described below.
 
 ## <a name="utteranceopts-option"></a>`model.utteranceOpts` Configuration ##
 
-<div class="infusion-docs-note"><strong>Note:</strong> If needed, please read the <a href="ComponentConfigurationOptions.md">Component Configuration Options</a> document for a full description of infusion component options.</div>
+<div class="infusion-docs-note"><strong>Note:</strong> if needed, please read the <a href="ComponentConfigurationOptions.md">Component Configuration Options</a> document for a full description of infusion component options.</div>
 
 Configuration of the Text To Speech component can be done through `model.utteranceOpts`. This model path is a Javascript object that contains attributes that users can use to define the behaviour of the [SpeechSynthesisUtterance instance](https://dvcs.w3.org/hg/speech-api/raw-file/tip/webspeechapi.html#utterance-attributes) (a part of the web speech API that the Text To Speech component interacts with).
 
-<div class="infusion-docs-note"><strong>Note:</strong> Not all speech synthesizers support all these attributes and some may take different ranges.</div>
+<div class="infusion-docs-note"><strong>Note:</strong> not all speech synthesizers support all these attributes and some may take different ranges.</div>
 
 These attributes include:
 
@@ -395,7 +395,7 @@ These attributes include:
             <th>Description</th>
             <td>
                 The <code>text</code> attribute allows you to set the text that you wish to be spoken.
-                <div class="infusion-docs-note"><strong>Note:</strong>Be careful with this attribute as it will override any text that was previously passed.</div>                
+                <div class="infusion-docs-note"><strong>Note:</strong> be careful with this attribute as it will override any text that was previously passed.</div>                
             </td>
         </tr>
         <tr>
--- a/src/documents/TextToSpeechAPI.md
+++ b/src/documents/TextToSpeechAPI.md
@@ -4,10 +4,10 @@ layout: default
 category: Components
 ---
 
-The **Text To Speech** component uses [Web Speech Synthesis API](https://dvcs.w3.org/hg/speech-api/raw-file/tip/speechapi.html#tts-section) to queue up and read texts.
+The **Text To Speech** component uses [Web Speech Synthesis API](https://dvcs.w3.org/hg/speech-api/raw-file/tip/webspeechapi.html#tts-section) to queue up and read texts.
 
 ## Browser Support ##
-The Text To Speech component can be used in browsers that support [Web Speech Synthesis API](https://dvcs.w3.org/hg/speech-api/raw-file/tip/speechapi.html#tts-section). At the time of writing, July 26 2016, these browsers include:
+The Text To Speech component can be used in browsers that support [Web Speech Synthesis API](https://dvcs.w3.org/hg/speech-api/raw-file/tip/webspeechapi.html#tts-section). At the time of writing, July 26 2016, these browsers include:
 * Chrome 31+
 * Chrome for Android 40+
 * Safari 7.1+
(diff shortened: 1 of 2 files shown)


waharnum:
Tapping @jobara who I talked about this issue with this morning.

src/documents/TextToSpeechAPI.md
```diff
@@ -37,9 +39,9 @@ Use the following function to create a Text To Speech component:
             <th>Parameters</th>
             <td>
                 <dl>
-                    <dt>options</dt>
+                    <dt>model.utteranceOpts</dt>
                     <dd>
-                        An optional data structure that configures the Text To Speech component, as described below.
+                        An optional data structure supplied to the component model that configures the Text To Speech component behaviour, as described below. As part of the component model, can be managed through the [ChangeApplier API](ChangeApplierAPI.md) to assist in coordination with other components.
```
jobara: Could you add a link to the model.utteranceOpts section?

jobara: Never mind, I see there is a section for the link in the table.

src/documents/TextToSpeechAPI.md
```diff
@@ -65,7 +69,7 @@ var tts = fluid.textToSpeech({
         <tr>
             <th>See also</th>
             <td>
-                <a href="#utteranceopts-option">Utteranceopts Option</a>
+                <a href="#utteranceopts-option">`model.utteranceOpts` Options</a>
```
jobara: This section is HTML. I think you meant to use the "`" for code styling. You'll need to wrap it in a `<code>` element instead.

src/documents/TextToSpeechAPI.md
```diff
@@ -372,11 +376,11 @@ The events fired by the Text To Speech component are described below.
     </tbody>
 </table>
 
-## utteranceOpts` Option ##
+## <a name="utteranceopts-option"></a>`model.utteranceOpts` Configuration ##
 
 <div class="infusion-docs-note"><strong>Note:</strong> If needed, please read the [Component Configuration Options](ComponentConfigurationOptions.md) document for a full description of infusion component options.</div>
```
jobara: Could you update the syntax used for the link here. It's in an HTML block so needs to use `<a>`?

src/documents/TextToSpeechAPI.md
```diff
@@ -449,15 +457,14 @@ fluid.textToSpeech({
...
 <table>
     <tbody>
         <tr>
             <th>Description</th>
             <td>
-                The <code>voiceURI</code> attribute is a string that specifies the speech synthesis voice and the location of the speech synthesis service that the web application wishes to use. Calling the <a href="#getvoices">getVoices</a> method returns an array of all available voices, from which you can get the value of the <code>voiceURI</code> attribute.<br />
-                <strong>Note</strong>: In Chrome and Safari, the <code>voiceURI</code> attribute is named <code>voice</code> instead.
+                The <code>voice</code> attribute must be a <a href="https://dvcs.w3.org/hg/speech-api/raw-file/tip/speechapi.html#speechsynthesisvoice">`SpeechSynthesisVoice`</a> object that specifies the speech synthesis voice that the web application wishes to use. Calling the <a href="#getvoices">getVoices</a> method returns an array of all available voices, from which you can select a valid `SpeechSynthesisVoice`.
```
jobara: I think you meant to use `<code>` instead of the "`".

src/documents/TextToSpeechAPI.md
```diff
@@ -449,15 +457,14 @@ fluid.textToSpeech({
     </tbody>
 </table>
 
-### voiceURI ###
-
+### voice ###
+<!-- See the errata at https://dvcs.w3.org/hg/speech-api/raw-file/tip/speechapi-errata.html - SpeechSynthesisUtterance must be passed a SpeechSynthesisVoice object to configure the voice under the latest API -->
```
jobara: I think we should either remove this HTML comment or add it as a note that is visible to the reader.

src/documents/TextToSpeechAPI.md
```diff
@@ -471,9 +478,16 @@ fluid.textToSpeech({
             <td>
 <pre>
 <code>
+
+// Assuming an existing textToSpeech component
+var availableVoices = textToSpeech.getVoices();
+var voiceToUse = availableVoices[0];
```
jobara: This part of the example seems strange as you are required to have a pre-existing textToSpeech component in order to configure another component. It may be better to just use the web speech api directly to access the voices and set the `voiceToUse` variable.

jobara: We may want to include a JIRA number related to modifying the options to take in a voice identifier.

src/documents/TextToSpeechAPI.md
waharnum: I don't see an output difference between the Markdown-style link vs. an HTML one - is this for the sake of consistency in the source document?

src/documents/TextToSpeechAPI.md
waharnum: I think some of this comment got eaten by an unescaped \`code-format backtick\`, so I'm not clear on it - it's already wrapped in backticks?

src/documents/TextToSpeechAPI.md
waharnum: @jobara I created a JIRA here: https://issues.fluidproject.org/browse/FLUID-5939

waharnum:
@jobara - I think I hit all the points of feedback on this.

src/documents/TextToSpeechAPI.md
```diff
@@ -7,11 +7,13 @@ category: Components
...
-The Text To Speech component can be used in browsers that support [Web Speech Synthesis API](https://dvcs.w3.org/hg/speech-api/raw-file/tip/speechapi.html#tts-section). At the time of writing, Mar 4 2015, these browsers include:
+The Text To Speech component can be used in browsers that support [Web Speech Synthesis API](https://dvcs.w3.org/hg/speech-api/raw-file/tip/speechapi.html#tts-section). At the time of writing, July 26 2016, these browsers include:
 * Chrome 31+
 * Chrome for Android 40+
 * Safari 7.1+
 * iOS Safari 7.1+
+* MS Edge 14 (in Preview Release)
+* Firefox 48+ (using the `media.webspeech.synth.enabled` about:config option)
 
 <div class="infusion-docs-note"><strong>Note:</strong> Find the latest browser support data for Web Speech Synthesis API from [caniuse.com](http://caniuse.com/#feat=web-speech).</div>
```
jobara: Another spot where a link is specified in markdown syntax but needs to use HTML.

src/documents/TextToSpeechAPI.md
jobara: The link for the change applier added here needs to be specified using HTML not MD.

src/documents/TextToSpeechAPI.md
```diff
@@ -449,15 +458,15 @@ fluid.textToSpeech({
...
-                <strong>Note</strong>: In Chrome and Safari, the <code>voiceURI</code> attribute is named <code>voice</code> instead.
+                <p>The <code>voice</code> attribute must be a <a href="https://dvcs.w3.org/hg/speech-api/raw-file/tip/webspeechapi.html#speechsynthesisvoice"><code>SpeechSynthesisVoice</code></a> object that specifies the speech synthesis voice that the web application wishes to use. Calling the <a href="#getvoices">getVoices</a> method returns an array of all available voices, from which you can select a valid <code>SpeechSynthesisVoice</code>, or you can call the <code><a href="https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesis/getVoices">SpeechSynthesis.getVoices()</a></code> function of the browser directly.</p>     
+                <div class="infusion-docs-note">Note that in some browsers (such as Chrome), the voice list is populated after the page is loaded, and you may need to wait for the <a href="https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesis/onvoiceschanged">onvoiceschanged</a> event to get a full list.</div>
```
jobara: The links to `SpeechSynthesis.getVoices()` and `onvoiceschanged` point to MDN instead of a W3 doc. Is that intentional?

waharnum: It was originally because those were up to date and the W3 docs weren't, but now that we're linking to the W3C "Unofficial Draft" (which matches the implemented browser APIs), I'll update them for consistency.

waharnum:
Do you think this one is good to go in, @jobara?

src/documents/TextToSpeechAPI.md
```diff
@@ -390,7 +395,7 @@ These attributes include:
             <th>Description</th>
             <td>
                 The <code>text</code> attribute allows you to set the text that you wish to be spoken.
-                <strong>Note</strong>: Be careful with this attribute as it will override any text that was previously passed.
+                <div class="infusion-docs-note"><strong>Note:</strong>Be careful with this attribute as it will override any text that was previously passed.</div>
```
jobara: Missing a space between `</strong>` tag and "Be"

src/documents/TextToSpeechAPI.md
```diff
@@ -449,15 +458,15 @@ fluid.textToSpeech({
...
-                <strong>Note</strong>: In Chrome and Safari, the <code>voiceURI</code> attribute is named <code>voice</code> instead.
+                The <code>voice</code> attribute must be a <a href="https://dvcs.w3.org/hg/speech-api/raw-file/tip/webspeechapi.html#speechsynthesisvoice"><code>SpeechSynthesisVoice</code></a> object that specifies the speech synthesis voice that the web application wishes to use. Calling the <a href="#getvoices">getVoices</a> method returns an array of all available voices, from which you can select a valid <code>SpeechSynthesisVoice</code>, or you can call the <code><a href="https://dvcs.w3.org/hg/speech-api/raw-file/tip/webspeechapi.html#dfn-ttsgetvoices">SpeechSynthesis.getVoices()</a></code> function of the browser directly.
+                <div class="infusion-docs-note"><strong>Note:</strong> in some browsers (such as Chrome), the voice list is populated after the page is loaded, and you may need to wait for the <a href="https://dvcs.w3.org/hg/speech-api/raw-file/tip/webspeechapi.html#dfn-ttsonvoiceschanged">voiceschanged</a> event to get a full list.</div>
```
jobara: Should "in" be capitalized here?

jobara:
@waharnum I'm noticing that the padding around the "note" in the tables has a large gap at the bottom but is right up against the text at the top. Do you thin we should balance out that spacing in the table?

screen shot 2016-08-22 at 2 58 11 pm

jobara:
merged at cbfe7f487943c1f49078cd6a81437ffa755c1a9d

PR #95, opened 2016-08-18
FLUID-5943: tutorial for a minimal footprint prefs editor

jobara:
Discusses using the full page prefs editor with lazy loading of the settings store and page enhancer.

https://issues.fluidproject.org/browse/FLUID-5943


jobara:
@colinbdclark could you please review the documentation?

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -0,0 +1,133 @@
...
+            gradeNames: ["fluid.prefs.starterFullPreviewTemplateLoader"]
+        }
+    });
+```
+
+### Lazy Loading ###
+
+With a Full Page Preferences Editor we've moved editing preferences off of each of the content pages; however, these pages are still required to enact/apply the preferences set by the user. Therefore, we still need to instantiate the [Settings Store](../SettingsStore.md) and [Page Enhancer](../Enactors.md) on each page. However, we only need these if a user has actually adjusted their preferences. In a default implementation of the Preferences Framework, preferences are stored in a browser cookie. The simple solution is to check for the presence of this cookie, and only load and instantiate the Settings Store and Page Enhancer if it's found.
+
+The following example makes use the functions provided by [loadScripts.js](../examples/loadScripts.js) to check the cookie and lazy load the required scripts:
```
amb26: Very peculiar that this is in "examples" within the docs and not within our core implementation - do we support this code or not? Doesn't the core implementation now already achieve lazy loading?
If we are recommending this code as part of a core user implementation strategy, it should meet the same standards as all the rest of our code - e.g. have test cases

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -0,0 +1,133 @@
...
+        },
+        templateLoader: {
+            gradeNames: ["fluid.prefs.starterFullPreviewTemplateLoader"]
+        }
+    });
+```
+
+### Lazy Loading ###
+
+With a Full Page Preferences Editor we've moved editing preferences off of each of the content pages; however, these pages are still required to enact/apply the preferences set by the user. Therefore, we still need to instantiate the [Settings Store](../SettingsStore.md) and [Page Enhancer](../Enactors.md) on each page. However, we only need these if a user has actually adjusted their preferences. In a default implementation of the Preferences Framework, preferences are stored in a browser cookie. The simple solution is to check for the presence of this cookie, and only load and instantiate the Settings Store and Page Enhancer if it's found.
```
amb26: Simply "off each" is enough

src/static/examples/loadScripts.js
```diff
@@ -0,0 +1,27 @@
+var fluid_load = fluid_load || {};
```
amb26: Why not "fluid"?

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -0,0 +1,133 @@
...
+        uiEnhancer: {
+            gradeNames: ["fluid.uiEnhancer.starterEnactors"],
+            tocTemplate: "path/to/TocTemplate"
+        }
+    });
+
+    /**
+     * Initialize Full Page preferences editor
+     */
+    fluid.prefs.fullPreview(".container", {
```
amb26: Shouldn't this example conform to all of our recommendations? E.g. that class name selectors are properly scoped,e tc.

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -0,0 +1,133 @@
...
+        gradeNames: ["fluid.prefs.initialModel.starter"],
+        // Tell PrefsEditor where to find all the templates, relative to this
+        // file
+        terms: {
+            templatePrefix: "path/to/templates",
```
amb26: Why not supply a fully working example, rather than just some placeholders? That way it would at least be possible to get some assurances that these docs are correct.

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -0,0 +1,133 @@
+---
+title: Preferences Framework: Minimal Footprint
+layout: default
+category: Tutorials
+---
+
+Website performance can be split into several buckets: backend, over the wire, and front end. It is important to pay attention to all of these areas when devising your optimization strategy. This document will focus on using the [Preferences Framework](../PreferencesFramework.md) in a situation where optimization is critically important; however, it will not go into detail to cover all of the potential optimizations that can be employed on your site.
```
amb26: This paragraph is self-contradictory. In the first sentence you explain that there are several "buckets" - in the second sentence you simply talk of "optimization" without distinguishing between buckets. "critically important" is pretty vague

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -0,0 +1,133 @@
...
+     * "starter" adjusters and enactors should be used. These correspond to the
+     * same set of adjusters and enactors used in a typical UI Options
+     * configuration. However, a custom set of adjusters and enactors could be
+     * configured here.
+     */
+
+    /**
+     * Initialize PrefsEditor global settings store.
+     */
+    fluid.prefs.globalSettingsStore();
```
amb26: Mention what this does and what its default strategy is

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
jobara: It's been a while, so it's a bit hard to remember, but I'll try to explain. This was just provided as an example of how one could address an extreme case of lazy loading where all of the dependencies are lazy loaded. That is, Infusion itself does not even appear on the page until the UI Enhancer is required. It's not necessarily the recommended case, I'd say the UIO Lazy loading docs cover what we'd typically recommend for lazy loading, but this covers the case where someone wishes to have virtually no impact from the Prefs Framework on their load times. To that end, the code is plain javascript and doesn't make use of anything from Infusion, and is really just one possible solution.

src/static/examples/loadScripts.js
amb26: I don't think there is any implication here. The code is under our definition and I don't see any reason to take up a separate top-level namespace - there's nothing about the definition of the "fluid" namespace that implies how complete its contents should be. In fact in future we will positively promote the use of "fluid bootstrappers" which enable low-powered clients get access to framework facilities cheaply without necessarily loading, for example, all of Fluid IoC.

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
amb26: All the same, I would say that if we promote this as a solution, we should curate the code in accordance with our values.

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
jobara: What do you mean by working example? I understand that you don't want "path/to/templates" but not sure what you want that replaced with. E.g. do you want the templates actually included in the docs or just a feasible path if they were to download infusion?

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
jobara: > All the same, I would say that if we promote this as a solution, we should curate the code in accordance with our values.

I take this to mean that you still believe this should be included in Infusion. Where would you suggest it go, prefs framework, core, somewhere else? loadScripts.js is fairly general and could be included anywhere. Also are we sure that we really want to be endorsing this type of script loading in Infusion? Should we at least mark it with the unsupported api tag?

jobara: Also, requiring this be moved into Infusion means that this PR is blocking the Infusion 2.0 release and not just the Docs. So we'll need to decide, implement, and test this as soon as possible.

jobara:
@amb26 i addressed some of your comments and have left some further questions for you as well. Thanks.

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
amb26: We need to supply a complete "ready to go" example, since people will work (as is typical) by copying it. Things like path/to/templates leave too many issues open to question, e.g. from what directory this path is relative to, etc.

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
amb26: > Also are we sure that we really want to be endorsing this type of script loading in Infusion

I would say if we are providing this code in our documentation repository, we are endorsing it - which implies we should support it. We can depend that any code that we write in such a place will immediately be cut and pasted into any number of user projects, so we had better be sure that it conforms to a best practice that we believe in. If we don't think we can support the use of it, we should remove it from our official framework docs and perhaps put it somewhere else - maybe a wiki page or simply a mailing list posting. Another good place would be our "examples" directory - https://github.com/fluid-project/infusion/tree/master/examples/framework/preferences - my suggestion is that it should be packaged as a suitably commented working sample there. At least in that case it could undergo some form of QA, however unofficial.

jobara: I talked with @michelled in the [fluid-work irc channel](https://botbot.me/freenode/fluid-work/2016-11-21/?msg=76804836&page=1) today. We've decide to move the code used here to the examples. I'll file a separate PR against Infusion for that, and then come back to this PR.

jobara:
Example code supplied in PR https://github.com/fluid-project/infusion/pull/789 to Infusion

jobara:
@amb26 ready for more review

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -0,0 +1,186 @@
...
+        },
+        templateLoader: {
+            gradeNames: ["fluid.prefs.starterFullPreviewTemplateLoader"]
+        },
+        prefsEditor: {
+            gradeNames: ["fluid.prefs.starterPanels", "fluid.prefs.uiEnhancerRelay"],
+            listeners: {
+                // Tells the PrefsEditor where to redirect to if the user cancels the operation.
+                // In this case, it goes back one step in the browser's history.
+                onCancel: {
```
amb26: This is pretty cool, isn't it : P

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -0,0 +1,186 @@
...
+The following example instantiates a Full Page Preferences Editor with a preview window.
+The code consists of the following three parts:
+
+1. `fluid.prefs.globalSettingsStore` - Responsible for storing/retrieving preferences. By default it uses the `fluid.prefs.cookieStore` which uses a browser cookie for storage.
+2. `fluid.pageEnhancer` - Initializes the UI Enhancer for the page. The UI Enhancer is responsible for applying the preferences to the page.
+3. `fluid.prefs.fullPreview` - The Full Page prefs editor, including a preview.
+
+(see: [fullPage.html](https://github.com/fluid-project/infusion/blob/master/examples/framework/preferences/minimalFootprint/fullPage.html))
+
+<div class="infusion-docs-note"><strong>Note:</strong> The code snipped below can be loaded in through a JavaScript file or added directly to the HTML in a <code>&lt;script&gt;</code> tag.</div>
```
amb26: snipped -> snippet?

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -0,0 +1,186 @@
...
+        // The UI Enhancer's Table of Contents uses an HTML template. This tells the component
+        // where to find that template.
+        tocTemplate: "../../../../src/components/tableOfContents/html/TableOfContents.html"
+    }
+});
+```
+
+## Caveats ##
+
+While the strategy outlined above will improve page load performance, it is not without its draw backs.
```
amb26: draw backs -> drawbacks

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -0,0 +1,186 @@
...
+        },
+        preview: {
+            templateUrl: "html/prefsEditorPreview.html"
+        }
+    });
+```
+
+### Lazy Loading ###
+
+With a Full Page Preferences Editor we've moved editing preferences off of the content pages; however, these pages are still required to enact/apply the preferences set by the user. Therefore, we still need to instantiate the [Settings Store](../SettingsStore.md) and [Page Enhancer](../Enactors.md) on each page. However, we only need these if a user has actually adjusted their preferences. In a default implementation of the Preferences Framework, preferences are stored in a browser cookie. The simple solution is to check for the presence of this cookie, and only load and instantiate the Settings Store and Page Enhancer if it's found.
```
amb26: "off of" has returned in the previous commit

jobara:
@amb26 ready for more review.

PR #98, opened 2016-09-01
FLUID-5370: Documenting token removals

jobara:
Related to PR https://github.com/fluid-project/infusion/pull/744

https://issues.fluidproject.org/browse/FLUID-5370


jobara:
This was discussed at the Community Meeting on Sept 14, 2016. It was determined that the old values were still useful to integrators and that all relevant information that could be tokenized and used within a string template should be modelled and passed into the string template function. This has been captured in [FLUID-5957](https://issues.fluidproject.org/browse/FLUID-5957)

PR #107, opened 2016-11-23
FLUID-5971: Docs update...

the-t-in-rtf:
Adds documentation for the transformations added in [FLUID-5791](https://issues.fluidproject.org/browse/FLUID-5791).


the-t-in-rtf:
@jobara, as you have been handling the code PR, do you want to field this one as well?

the-t-in-rtf:
@amb26 and @jobara, the associated code pull has been merged, this still needs to be looked at (I can't assign it myself).

the-t-in-rtf:
@amb26, this now reflects the wording we agreed on for the date/string transforms in the other PR.

src/documents/ModelTransformationAPI.md
```diff
@@ -3217,3 +3224,386 @@ transform: {
...
+</code></pre></td>
+</tr>
+</tbody>
+</table>
+
+### fluid.transforms.stringToDate
+
+**Type:** standardTransformFunction
+
+**Description:** Transform a string to a date using the Date constructor.  Accepts (among other things) the date and dateTime values returned by HTML5 date and dateTime inputs. Note: This function allows you to create Date objects from an ISO 8601 string such as `2017-01-23T08:51:25.891Z`.  It is intended to provide a consistent mechanism for recreating Date objects stored as strings.  Although the framework currently works as expected with Date objects stored in the model, this is very likely to change.  If you are working with Date objects in your model, your best option for ensuring your code continues to work in the future is to handle serialisation and deserialisation yourself, for example, by using this transform and one of its inverse transforms, `fluid.transforms.dateToString` or `fluid.transforms.dateTimeToString`.  See [the Infusion documentation](http://docs.fluidproject.org/infusion/development/FrameworkConcepts.html#model-objects) for details about supported model values.
```
amb26: I don't think that what you mean by "handle serialisation and deserialisation yourself" is very clear. The principal risk that people face at the moment is that i) users of Date objects often tend to treat them as mutable (given they have a mutable API) whereas ii) the framework simply ships them around as if they were immutable - that is, the same Date object will end up embedded in multiple model trees. Assuming that the user does not try to modify the particular Date objects that they find or put within models, everything will be fine. Given this, "works as expected" is a bit of a hedge. I'd be happy to just take on this wording and edit it myself if you're ok with that, since I know we have had a few passes at it already.

the-t-in-rtf: @amb26, that would be good.  This is exactly the wording that was merged in the JSDocs for the infusion pull, so it would be good to close the loop and make sure that's also accurate once you have wording you're happy with.

PR #110, opened 2016-12-12
FLUID-6093: Upgrade to use Infusion 2.0

jobara:
https://issues.fluidproject.org/browse/FLUID-6093


src/static/lib/infusion/Version.md
```diff
@@ -1,19 +1,8 @@
...
-    grunt custom --source=true --include="uiOptions" --name="uio"
-```
-
-The file src/framework/preferences/html/SeparatedPanelPrefsEditorFrame.html was modified to
-reference infusion-uio.js instead of individual js files.
-
-The following directories and files were stripped out of the build, since they contain
-code that is included in the infusion-uio.js file or is not required by this plugin:
+The following directories and files were stripped out of the build, because they contain
+code that is included in the infusion-uiOptions.js file or is not required:
```
jhung: rephrase to: "The following directories and files were removed from the build, because they are already included in infusion-uiOptions.js file, or are no longer needed:"

jobara:
@jhung made the change you recommended. Ready for more review.

PR #112, opened 2016-12-15
NOJIRA: small correction to documentation

waharnum:
This corrects a small (presumed copy-paste originating) error in the documentation that we found during the pair programming community meeting.

`fluid.values` was stating it returned the keys of a supplied object as an array; as might be expected, it in fact returns the values, and `fluid.keys` is the equivalent key-listing function. 💯


jobara:
merged at 1f2be2f7adb3469244a9ac5d9affd7c8f0bcd8da

PR #114, opened 2017-01-16
FLUID-6112: add a Developer Introduction to core concepts of Infusion

waharnum:
JIRA: https://issues.fluidproject.org/browse/FLUID-6112

This is a document for Infusion 2.0 in the style of https://wiki.fluidproject.org/display/docs/Developer+Introduction+to+Infusion+Framework; it stays away from the pre-built components entirely (those might be covered in another, similar document) in favour of explaining some of the most important core framework concepts. It assumes a developer has some experience with Javascript development and jQuery.

I've used CodePen at each stage for a live code example of what's being demonstrated.

Tear it up, because we want this to be useful for people who haven't worked with the framework before. :)


src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+
+Infusion also makes extensive use of the programming concept of [Inversion of Control](/infusion/development/FrameworkConcepts.md#ioc) (IoC), a technique for organizing component dependencies and references in a distributed, flexible manner. IoC is used in many ways throughout Infusion, but an important initial concept to grasp is the use of [IoC References](/infusion/development/IoCReferences.md) when configuring components, and the concept of `{that}`.
+
+IoC references allow us to refer to other parts of a component object (or another component entirely) in a declarative, context-specific manner, with `{that}` standing in for the current component configuration.
+
+We can use a listener definition, the `onCreate` lifecycle event and IoC `{that]` to have a component say hello when it's ready, rather than needing to be manually called:
```
michelled: typo on this line '{that]'

the-t-in-rtf:
As the display of mixed HTML/markdown paragraphs is broken on GitHub, I am reviewing in docpad.  My first question is, where will this live?  It's not linked into any of the top level categories at the moment.

the-t-in-rtf:
I'm getting the idea that this is meant to be a tutorial?  It should be linked there.  In any case, IMO it would be better to break this down into separate pages with a shared table of contents.  For example, I think there's too big a gap between the first principles introduced and the summary.  If you were to restructure into smaller pages, each page would hold part of the summary, and you could shorten each point in the overall summary and link to the section that demonstrates the principle.

If you break it out into sections, you also have more room to link to background information on general concepts we are applying (I will comment inline with examples).

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+# Introduction
+
+Here we will boil down some of the basic concepts of Infusion for developers who may be interested in it, but uncertain of where to start. There's a lot going on in the framework, but grasping some core ideas helps a great deal in moving forward and learning.
+
+This introduction summarizes some topics (and leaves out others) that are gone into at length in the longer [Framework Concepts](/infusion/development/FrameworkConcepts.md) document; developers interested in greater detail about the guiding philosophy behind Infusion should consult that document as well.
+
+The introduction assumes you are familiar with the basics of [JavaScript](https://developer.mozilla.org/en-US/docs/Learn/JavaScript) development, and with using a browser's [developer tools](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_are_browser_developer_tools). You should also have some experience of the [jQuery](https://jquery.com/) JavaScript library that Infusion is built on top of.
+
+We will present an increasingly complicated "Hello, World!" implementation to demonstrate the core concepts, with live examples throughout using the [CodePen](http://codepen.io/) service.
```
the-t-in-rtf: "complex" seems better than "complicated" here.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+        speakHello: {
+            type: "fluid.helloWorld.speakHello",
+        }
+    }
+});
+```
+
+### Summing Up
+
+This elaborate "Hello, World!" implementation has slowly introduced some of Infusion's core features. To summarize:
```
the-t-in-rtf: Perhaps "moderately complex" instead of "elaborate" would be better here.  Also, "introduces" rather than "has slowly introduced" seems better.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+
+* We implement our programs by designing [components](#components) that work together to provide the needed functionality. Components are Javascript objects adhering to certain syntax conventions that are created using the `fluid.defaults` function. While components have defaults, any of these defaults can be overriden at the time a specific instance of the component is created.
```
the-t-in-rtf: As I mentioned in my global comment, if you break things down by sections, you can expand on each concept within the section, and then use a condensed version here, linked to the expanded definition.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+
+ * [`fluid.component`](/infusion/development/ComponentConfigurationOptions.md#options-supported-by-all-components-grades), the simplest type with support for events and public function definition (called [invokers](/infusion/development/Invokers.md) in Infusion).
+ * [`fluid.modelComponent`](/infusion/development/ComponentConfigurationOptions.md#model-components), which adds support for mutable state, model events, and state coordination and relay between different components.
+ * [`fluid.viewComponent`](/infusion/development/ComponentConfigurationOptions.md#view-components), which supplement model components with functionality to bind to the DOM of a web page.
+
+Component-based development emphasizes declarative configuration, loose coupling and flexible hierarchies.
+
+A new Infusion component is defined using the `fluid.default` function and a very basic "Hello, World!" component might look like this:
+
+<div class="infusion-docs-note">You can check out the [Live Example](http://codepen.io/waharnum/pen/oBYvwx?editors=1111) of the code below on [CodePen](http://codepen.io/)</div>
```
the-t-in-rtf: None of the codepen examples worked for me when I tested, seems to be a problem on their end.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+            "this": "{that}.dom.messageArea",
+            // Calls the 'html' function to replace the HTML at the node
+            "method": "html",
+            "args": ["{that}.model.message"]
+        }
+    }
+});
+```
+
+Then, from the console, we'll use the ChangeApplier to change the model; notice how the modelListener we defined responds to the change and updates the message again:
```
the-t-in-rtf: Highlight ChangeApplier in backticks.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+```
+
+After definition, instances of the component can be created by calling the component's name as a function, with the option to pass in further configuration options as an argument:
+
+```
+var helloWorld = fluid.helloWorld({});
+```
+
+Right now this component doesn't do anything new, but we will evolve its definition throughout this introduction to demonstrate further core concepts of the framework.
+
```
the-t-in-rtf: As an example, this would be a natural point for a section break.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+            "method": "log",
+            "args": ["Hello, World!"]
+        }
+    }
+});
+```
+
+### Models and Model Components
+
+Mutable data is expected to be stored on a component's [model](/infusion/development/FrameworkConcepts.md#model-objects). Component models in Infusion are altered through the [ChangeApplier](/infusion/development/ChangeApplier.md) functionality, which works to:
```
the-t-in-rtf: This needs to be fleshed out more.  We obviously use non-model mutable data (member variables) all the time.  Also, assume they might not have familiarity with MVC or what's good to put in a model, a link and a sentence summary would be helpful.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+    }
+});
+```
+
+Now we can create an instance of the component and call the invoker as a function to say hello:
+
+```
+var helloWorld = fluid.helloWorld({});
+helloWorld.sayHello();
+```
```
the-t-in-rtf: At the risk of making this too complex, IMO it would be good to demonstrate how an invoker can be overridden.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+                }
+            }
+        }
+    }
+});
+```
+
+### Restructuring Infusion Components
+
+Infusion's configuration-oriented components make it easier to restructure code, especially as component configuration becomes unwieldy. In the example below, we extract the two "say hello" components into separate component definitions from the main component, then include them as subcomponents of the main component. We've also added a listener to the main component to announce (once) when it is finished creation.
```
the-t-in-rtf: "it is finished creation" seems odd.  Perhaps "its creation is complete"?

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+            type: "fluid.helloWorld.consoleHello"
+        },
+        consoleLogger: {
+            type: "fluid.helloWorld.displayHello",
+        }
+    }
+});
+```
+
+### Using Invokers for Polymorphic Behaviour and Refactoring to Reuse Configuration
```
the-t-in-rtf: If you're going to use the word "polymorphic" multiple times in the doc, you should really have more than a deep dive link into our docs on the subject.  A sentence summary of what we mean and a link to a WikiPedia article on the topic would be nice.

the-t-in-rtf: Or perhaps just add another link to "our" coverage of the subject here, you have one near the other references.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+        },
+    }
+});
+
+fluid.defaults("fluid.helloWorld.displayHello", {
+    // This component has all of the characteristics of sayHello,
+    // except for its implementation in the invoker
+    gradeNames: ["fluid.helloWorld.sayHello"],
+    invokers: {
+        sayHello: {
```
the-t-in-rtf: I commented earlier about demonstrating overriding.  Mentioning it at the end of the "invokers" section and linking to this section would be a good way to handle that.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+* Model components can use [model listeners](#listening-to-model-changes) to respond to changes in model state.
+* Components can include [subcomponents](#subcomponents-and-model-relaying), and use [model relaying](#subcomponents-and-model-relaying) to keep state synchronized between different components in larger designs. Many kinds of model relays are possible aside from two-way synchronization.
+* As program designs evolve, Infusion's configuration-oriented components make it easier to [restructure a design](#restructuring-infusion-components) by splitting out functionality into multiple components and wiring them together through IoC references.
+* When it becomes clear two components have similar behaviour, Infusion's design helps in [refactoring to share functionality](##using-invokers-for-polymorphic-behaviour-and-refactoring-to-reuse-configuration) through the use of invokers and base grades.
+* Infusion has strong supports for [multimodal implementations](#adding-further-ways-of-saying-hello) that allow programs to adapt themselves to new contexts for input and presentation.
```
the-t-in-rtf: The summary needs a bit of work.  As I've stated elsewhere, by this point you should be summarizing in much less detail, and linking back to the topic if people want to refresh their memory.

You should also really end with a "next steps" section, give them an idea of what they might do to continue learning from this point.  Even something as simple as encouraging them to join the community and linking to those docs would be a good start.

I'm also noticing that we don't really make it easy to close the loop.  We should end most tutorials with some idea of how to give feedback.  If we say it only in introductory material, people are likely to forget it if they have questions or find errors while reading the actual tutorial.

waharnum:
@the-t-in-rtf, thanks for all the comments - I think splitting it out into separate pages is entirely appropriate. I'd begun writing this as a short one-pager, and then it sort of evolved as I decided it needed to have code examples to illustrate the comments.

I'll be working on revisions in the next day or so once others have had a chance to comment.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
 
-_This page is still under construction. An archived version can be found on the
+<!-- _This page is still under construction. An archived version can be found on the
 [Developer Introduction to the Infusion Framework](http://wiki.fluidproject.org/display/docs/Developer+Introduction+to+Infusion+Framework)
-page of the [Infusion Documentation wiki](http://wiki.fluidproject.org/display/docs/Infusion+Documentation)._
+page of the [Infusion Documentation wiki](http://wiki.fluidproject.org/display/docs/Infusion+Documentation)._ -->
+
+# Introduction
+
+Here we will boil down some of the basic concepts of Infusion for developers who may be interested in it, but uncertain of where to start. There's a lot going on in the framework, but grasping some core ideas helps a great deal in moving forward and learning.
```
amb26: Really great work on this tutorial, WWAHHARNNUM!

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+    // gradeNames supplies an array of strings that name previously
+    // defined Infusion components
+    gradeNames: ["fluid.component"]
+});
+```
+
+After definition, instances of the component can be created by calling the component's name as a function, with the option to pass in further configuration options as an argument:
+
+```
+var helloWorld = fluid.helloWorld({});
```
amb26: You can leave out the {} since there are no options

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+
+We will present an increasingly complicated "Hello, World!" implementation to demonstrate the core concepts, with live examples throughout using the [CodePen](http://codepen.io/) service.
+
+## Infusion's Core Concepts
+
+### Components
+
+Programs using Infusion structure their data, functionality and interaction as various components defined using a common syntax.
+
+Components are [regular JavaScript objects that have certain characteristics](/infusion/development/UnderstandingInfusionComponents.md). Components can be freely combined together to form new components using the [grades system](/infusion/development/ComponentGrades.md), or organized into relationships with one another via [subcomponents](/infusion/development/SubcomponentDeclaration.md).
```
amb26: Not sure what it really adds to say that they are "regular JavaScript objects", but I guess it's the kind of thing that people say. You might want to say something, even at this early stage, about how they have their lifetime and scoping managed by the framework. Whilst you could say they are "a bit regular" they are not, for example, plain JSON-equivalent objects since (at least for the last couple of years) they have a non-default prototype supplied via Object.create().

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+
+ There are three base components to be aware of that any new components you create with Infusion will likely have at their root:
+
+ * [`fluid.component`](/infusion/development/ComponentConfigurationOptions.md#options-supported-by-all-components-grades), the simplest type with support for events and public function definition (called [invokers](/infusion/development/Invokers.md) in Infusion).
+ * [`fluid.modelComponent`](/infusion/development/ComponentConfigurationOptions.md#model-components), which adds support for mutable state, model events, and state coordination and relay between different components.
+ * [`fluid.viewComponent`](/infusion/development/ComponentConfigurationOptions.md#view-components), which supplement model components with functionality to bind to the DOM of a web page.
+
+Component-based development emphasizes declarative configuration, loose coupling and flexible hierarchies.
+
+A new Infusion component is defined using the `fluid.default` function and a very basic "Hello, World!" component might look like this:
```
amb26: typo, should be fluid.defaults

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
```diff
@@ -4,6 +4,595 @@ layout: default
...
+                        "args": ["{that}.model.message"]
+                }
+                }
+            }
+        }
+    }
+});
+```
+
+### Restructuring Infusion Components
```
amb26: Good example - since my first thought on reading the previous section was, "These definitions should be extracted as grades" : P

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
```diff
@@ -0,0 +1,598 @@
...
+
+ There are three base components to be aware of that any new components you create with Infusion will likely have at their root:
+
+ * [`fluid.component`](/infusion/development/ComponentConfigurationOptions.md#options-supported-by-all-components-grades), the simplest type with support for events and public function definition (called [invokers](/infusion/development/Invokers.md) in Infusion).
+ * [`fluid.modelComponent`](/infusion/development/ComponentConfigurationOptions.md#model-components), which adds support for mutable state, model events, and state coordination and relay between different components.
+ * [`fluid.viewComponent`](/infusion/development/ComponentConfigurationOptions.md#view-components), which supplement model components with functionality to bind to the DOM of a web page.
+
+Component-based development emphasizes declarative configuration, loose coupling and flexible hierarchies. The framework manages component lifecycle and scoping and provides many supports to allow components to work in collaboration with one another.
```
klown: @waharnum, in your presentation at the community meeting, you went into a little more detail about the "loosely coupled" and "flexible hierarchies", comparing infusion to frameworks that use a more rigid derivation hierarchy.  You might want to expand this written version in a similar way.

Having said that, however, a reason that infusion is flexible is because it's based on JavaScript, and JavaScript itself promotes flexible loose coupling.  That is, it's easy in JavaScript to use composition to create objects with the desired functionality, and, while it's possible to create class-like hierarchies in JavaScript, it's not as straight forward.

Infusion capitalizes on JavaScript's inherent abilities in this regard, whereas some other frameworks/toolkits emulate class-based approaches and are more rigid as a consequence.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
```diff
@@ -0,0 +1,598 @@
...
+    // gradeNames supplies an array of strings that name previously
+    // defined Infusion components
+    gradeNames: ["fluid.component"]
+});
+```
+
+After definition, instances of the component can be created by calling the component's name as a function, with the option to pass in further configuration options as an argument:
+
+```
+var helloWorld = fluid.helloWorld();
```
klown: In the previous version, an empty object was passed as an argument:
`var helloWorld = fluid.helloWorld({});`

I was going to comment that an empty object is not needed.  Still, I am unclear what is the best practice here.  Is there any difference between passing nothing and passing an empty object?  I should really test it and see.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
```diff
@@ -0,0 +1,598 @@
...
+<div class="infusion-docs-note">You can check out the [Live Example](http://codepen.io/waharnum/pen/MJbgEx?editors=1111) of the code below on [CodePen](http://codepen.io/)</div>
+
+```
+fluid.defaults("fluid.helloWorld", {
+    gradeNames: ["fluid.component"],
+    invokers: {
+        sayHello: {
+            // Configures this invoker to use the console object
+            "this": "console",
+            // Configures this invoker to use the log function of the console
```
klown: Pretending to be a novice JavaScript programmer here and asking intentionally naive questions:  The use of `"this"` is puzzling.  What is `"this"` in this context?  Why is it used?

What first pops into mind is that `"this"` refers to an instance of a `helloWorld` object created at run-time.  But that can't be right since `helloWorld` doesn't have a `console` member.  As it happens, in this case, `console` is a member of the `window` object in the browser, so the next question is: why not use `window` instead of `"this"`?

I've tried to come up with a concise transparent explanation of what `"this"` means here, but I'm stumped.

A related overall comment:
It would be useful to note explicitly that `gradenames`, `invokers`, `listeners`, etc. are reserved words in infusion, although I'm not sure that "reserved word" is the correct term here.  Perhaps, "behave like reserved words"?  Anyhow, these are fixed and have special meaning within both infusion and the `fluid.defaults()` function.  Similarly, within an  invoker block, `"this"`, `"method"`, `"args"`, "`func`", and "`funcName"` are reserved.  There are others like `"model"`, etc.

It would be useful to call out when the left side of the colon is defined by the framework (e.g., `invokers`), and when it's up the programmer (e.g., `sayHello`).

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
amb26: The result is exactly the same

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
amb26: "this" is used, because JavaScript function application is always implicitly bound to a variable whose name becomes "this" in the resulting scope, and needs to be explicitly nominated at the point of function application - see the signature of the builtin Function.apply: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply

We don't use "window" when invoking methods of "console" because this would cause the invocation to fail. The methods of the "console" object are implemented expecting that the value of "this" is bound to the console object itself. We only need to use the "this"-ist style when invoking code written by 3rd parties because all code written by Infusion natives ignore the value of "this", and so it is fine for this to be null or undefined.

Yes, I agree that the term "reserved word", imported from the world of programming languages, is misleading. It implies that the string has a special meaning *regardless of context* which is what this term implies in a programming language grammar. If there is any string which has this status within Infusion, it is only the string "expander" which almost universally throughout the framework announces that the following material will be evaluated by calling some function at instantiation time. Other than this, all the other strings, e.g. "this", "method", "gradeNames", etc. only have particular contextual meanings, for example when defining the body of an an invoker.

Yes, I agree that it would be useful to be clear when a structure has "free keys" and when it doesn't. We just need a reasonable term to describe "nonfree keys".

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
amb26: Yes, we should be careful in the statement "Component-based development emphasizes declarative configuration" since in general component-based development emphasizes exactly the opposite. It is our particular components, and their close relationship to JSON's free-form containment model, which promote this.

src/documents/to-do/DeveloperIntroductionToInfusionFramework.md
waharnum: In thinking about this and reading the Wikipedia article, I removed all mentions of it in favour of simply talking about invoker/function overriding.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
waharnum: Here's how I rephrased this:

```
Infusion's approach to component-based development emphasizes declarative configuration, loose coupling and flexible hierarchies. The framework manages component lifecycle and scoping and provides many supports to allow components to work in collaboration with one another.
```

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
waharnum: An existing phrasing I can find for this concept (in `C#`) is "contextual keywords": https://msdn.microsoft.com/en-us/library/x53a06bb.aspx#Anchor_0, which I think would be good enough for having a reasonable precise term to describe that a key such as `args` will be taken by the framework to have a particular meaning (and corresponding action taken by the framework to wire behaviours together) in certain contexts when encountered in parsing a configuration.

The Wikipedia entry (https://en.wikipedia.org/wiki/Reserved_word) notes the following:

> "A closely related and often conflated notion {to a reserved word} is a *keyword* which is a word with special meaning in a particular context."

waharnum:
This is ready for more review and should be viewable at http://www.alanharnum.ca/infusion-docs/infusion/development/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.html

Per feedback from various channels (this thread and the community meeting), I've made significant revisions including:

* Separating it into multiple linked sections
* Adding "Next Steps" material at the end
* Revising text and language, refactoring and adding comments to the code examples throughout to increase clarity and model good practice for component design 

I'd like to work towards getting this into a state where it can be released into the wild, so review towards any needed changes to the point where the community feels that can happen is appreciated.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-EventsAndInversionOfControl.md
```diff
@@ -0,0 +1,45 @@
...
+        // allow multiple listeners to be attached to the
+        // same event without collision
+        //
+        // On the right side, this configures the listener
+        // to use the sayHello invoker 
+        "onCreate.sayHello": "{that}.sayHello"
+    },
+    invokers: {
+        sayHello: {
+            "this": "console",
```
the-t-in-rtf: The invoker examples I've read thus far seem to use the "this-ist" form.  It seems more appropriate to use the ["standard" form](http://docs.fluidproject.org/infusion/development/Invokers.html#standard-invoker-binding-to-a-function-using-funcname-func) with perhaps an example of why someone might need to use the "this-ist" alternative.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-SubcomponentsAndModelRelaying.md
```diff
@@ -0,0 +1,79 @@
...
+                },
+                modelListeners: {
+                    "message": "{that}.displayHello"
+                },
+                invokers: {
+                    displayHello: {
+                        "this": "{helloWorld}.dom.messageArea",
+                        "method": "html",
+                        "args": ["{that}.model.message"]
+                }
```
the-t-in-rtf: This line should be indented.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-RestructuringComponents.md
```diff
@@ -0,0 +1,116 @@
...
+    components: {
+        consoleHello: {
+            type: "fluid.helloWorld.consoleHello",
+            options: {
+                model: {
+                    message: "{helloWorld}.model.message"
+                }
+            }
+        },
+        dispayHello: {
```
the-t-in-rtf: Typo here, "dispayHello".

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OverridingInvokersAndRefactoring.md
```diff
@@ -0,0 +1,101 @@
...
+<div class="infusion-docs-note">You can check out the [Live Example](http://codepen.io/waharnum/pen/bgBbOm?editors=1111) of the code below on [CodePen](http://codepen.io/)</div>
+
+```
+fluid.defaults("fluid.helloWorld.sayHello", {
+    gradeNames: ["fluid.modelComponent"],
+    model: {
+        message: "Hello, world!"
+    },
+    modelListeners: {
+        "message": "{that}.sayHello"
```
the-t-in-rtf: There is a mix of "quoted keys that need to be quoted", "quoted keys that do not need to be quoted", and "unquoted keys" throughout the  examples.

For consistency, I would suggest paring this down to just "keys that need to be quoted" and "unquoted keys".

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ExtendingDesignsWithExistingComponents.md
```diff
@@ -0,0 +1,120 @@
...
+Infusion's origins are in research about how to better implement accessible, inclusive systems. Many features are designed to support the easier transformation of data into other forms of representation, so it can be experienced by audiences with diverse needs.
+
+Our community often refers to this as *multimodal design* (in this context, a "modality" is "a particular mode in which something exists or is experienced or expressed"), and many characteristics of Infusion are about:
+
+* increasing the capacity of a system to have different representations of the same content, or to transform content into a form preferable to the end user
+* increasing the capacity of a system to have new forms of representation developed and added as needs arise for them
+
+We currently have a "Hello, World!" component that can say "hello!"" by printing to a web page and logging to the console. What if we wanted it to talk as well?
+
+We'll extend the code further using an existing Infusion [text to speech](/infusion/development/TextToSpeechAPI.html) component, which should work in a modern text-to-speech supporting browser:
```
the-t-in-rtf: The link here is broken.  Also, you use "text to speech" and "text-to-speech" in the same sentence.  My advice would be to pick one and make it consistent throughout all of these pages.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ExtendingDesignsWithExistingComponents.md
```diff
@@ -0,0 +1,120 @@
...
+    components: {
+        consoleHello: {
+            type: "fluid.helloWorld.consoleHello",
+            options: {
+                model: {
+                    message: "{helloWorld}.model.message"
+                }
+            }
+        },
+        dispayHello: {
```
the-t-in-rtf: Same "dispay" typo here.  Review all examples.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OverridingInvokersAndRefactoring.md
```diff
@@ -0,0 +1,101 @@
...
+    components: {
+        consoleHello: {
+            type: "fluid.helloWorld.consoleHello",
+            options: {
+                model: {
+                    message: "{helloWorld}.model.message"
+                }
+            }
+        },
+        dispayHello: {
```
the-t-in-rtf: Another typo on "dispay".

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ExtendingDesignsWithExistingComponents.md
```diff
@@ -0,0 +1,120 @@
...
+        sayHello: {
+            "this": "console",
+            "method": "log",
+            "args": ["{that}.model.message"]
+        },
+    }
+});
+
+fluid.defaults("fluid.helloWorld.displayHello", {
+    gradeNames: ["fluid.helloWorld.sayHello", "fluid.viewComponent"],
```
the-t-in-rtf: The introductory material on picking a grade doesn't mention inheritance.  You should probably mention that the order matters, with a simple example.  If it's still true that `fluid.viewComponent` needs to be the rightmost grade, that also needs to be mentioned.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ExtendingDesignsWithExistingComponents.md
the-t-in-rtf: Looks like the problem is with the leading slash.  Although this wouldn't be a problem when deploying to the root of a static site, it's better to have something that would also work in a subdirectory. As far as I can see, `../TextToSpeechAPI.html` would work here.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Developer Introduction to the Infusion Framework - Components
+layout: default
+category: Tutorials
+---
+
+Programs using Infusion structure their data, functionality and interaction as various components defined using a common syntax.
+
+Components are [JavaScript objects that have certain characteristics](/infusion/development/UnderstandingInfusionComponents.md). Components can be freely combined together to form new components using the [grades system](/infusion/development/ComponentGrades.md), or organized into relationships with one another via [subcomponents](/infusion/development/SubcomponentDeclaration.md).
```
the-t-in-rtf: All of these links are broken because they are deployed to a subdirectory.  Long-term, I can imagine us wanting to have draft docs, versioned docs, both of which would likely need to live in some kind of subpath rather than the root.  We also want it to be possible to view work-in-progress on GitHub pages and other static sites.  I'd suggest searching for '(/infusion/development' and replacing with '(../' with case-by-case review.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
```diff
@@ -0,0 +1,43 @@
...
+
+ There are three base components to be aware of that any new components you create with Infusion will likely have at their root:
+
+ * [`fluid.component`](/infusion/development/ComponentConfigurationOptions.md#options-supported-by-all-components-grades), the simplest type with support for events and public function definition (called [invokers](/infusion/development/Invokers.md) in Infusion).
```
the-t-in-rtf: I will stop pointing out the broken links from this point on, but wanted to comment that even though this link should work once we deploy, it should be possible to also test your links as part of the review.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-EventsAndInversionOfControl.md
waharnum: I've considered this myself, but I worry about modelling a less-than ideal practice of creating a free function that simply performs `console.log` on one argument. I followed this pattern myself in my early Infusion code (writing short functions that could be expressed in the "this-ist" style instead) and was typically asked to change it to "this-ist" style at review. The tutorial as-is requires no code that isn't expressed as a configuration tree due to the "this-ist" style of most of the invokers.

However, I think there would be some advantages to the approach you describe:

1) It introduces the concept of referring to free functions and the "standard" invoker form earlier.
2) It is probably somewhat easier to explain to a developer familiar with Javascript but not familiar with Infusion.
3) For the invokers dealing with selectors and the DOM, it is a chance to explain in more detail what the DOM binder does, contrast it to the jQuery-style of direct selector references in code and explain its value, and introduce the `locate` invoker that viewComponents have.
4) It delays the need to discuss the "this-ist" pattern, which I find somewhat hard to explain in the context of this tutorial

I'd be curious on the thoughts of others on this, since I don't think it has a clear answer.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-RestructuringComponents.md
waharnum: The phenomenon described at https://www.wired.com/2014/08/wuwt-typos/ in action. Fixed.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-EventsAndInversionOfControl.md
amb26: Typically what I do in these situations is to illustrate calling fluid.log rather than console.log, in order to avoid having to show the "this"-ist style. One could argue that this involves swapping one uncharacterised thing for another and so doesn't represent much improvement, but perhaps one could say that fluid.log represents a smaller unknown than the declarative invocation style that we're trying to explain.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ExtendingDesignsWithExistingComponents.md
waharnum: These should all be fixed now to fully relative links.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-EventsAndInversionOfControl.md
waharnum: Experimenting with using `fluid.log`, it seems necessary to do one of two things to use it as a substitute for `console.log` in this context...

1) Use the two-argument form: `[fluid.logLevel.IMPORTANT, "Hello, World!!"]`.
2) Call `fluid.setLogging(fluid.logLevel.INFO)` at some point before the call.

As much as I'd like to keep things simple and short as possible, this may simply require more expansion to be a really adequate explanation of what invokers "are" and the purpose they serve; I think it is probably important to comprehend both the Infunion-ist "that" (function references should be globally resolvable) from the reality of much of the rest of the Javascript world.

At the moment my feeling is that it is important to highlight three uses (as described at http://docs.fluidproject.org/infusion/development/Invokers.html):

1) Referring to free functions 
2) Referring to IoC resolved functions
3) "this"-ism

I'll continue to think on this one and revise accordingly.

amb26: Ok - another possibility is that you just write your own global function called "examples.log" that just calls console.log.

One global issue just struck me about this, which is that all your code is in the "fluid" namespace, which is not something we want to encourage in general user code - since all they are going to do is cut and paste this material and start working with it. The whole point of having the namespaces is to assist people from different communities to keep their work separate - so I suggest some different namespace for these examples such as examples.* or tutorial.*

colinbdclark: A general comment, which I hope isn't out of place or annoying:

In Flocking, years ago I stupidly made an example that uses the namespace <code>my.project</code> or something like that. Even today, I regularly see code in the wild where the developer didn't even bother to customize the namespace for their project. Most even reproduce the comment immediately above the <code>fluid.registerNamespace</code> call, which said "// Change this to something more appropriate for you project!" My guess is that 75% of the deployed Flocking applications out there are members of the <code>my.project</code> namespace. 

I think it's important to always design example code for the strange reality that there is no such thing as example code. As soon as it is published, examples will become "production code" for someone, and most likely for a lot of people.

waharnum: Both good points - perhaps something like the following:

1) Use a `fluidTutorial` namespace (it is a tutorial being produced within our community) for the new components, or perhaps `fluidTutorial.alan` (I suspect developers are less likely to simply retain a namespace with a person's actual name attached)

2) Explain why Infusion has the emphasis it does on namespacing - this could be added at the very start when talking about creating components

waharnum: All tutorial components are now namespaced to `fluidTutorial`.

waharnum:
Today's pushes should address the remaining outstanding comments:

1) Components now use `fluidTutorial` as their namespace instead of `fluid`. Some explanation of the importance of namespaces in Infusion has been added.

2) All `console.log` "this"-ist invokers were redone to use a free function. The "this"-ist style was retained for DOM manipulation using jQuery by the `viewComponent` example code, with an explanation of this in the view component section.

All live code samples have been correspondingly updated as well.

waharnum:
How do we feel about having this merged in in current state? Are there any outstanding issues that should be addressed?

waharnum:
Per last April's community meeting feedback, I've made the following revisions:

* Added a final section on transformation when doing model relay, using the straightforward "single input function as transform" case.
* Replaced 'shared state'-type wording with the broader and more accurate 'coordinated state'.
* Fixed a few small spelling / wording issues.
* Added additional explanation of what's happening in the section on subcomponents and model relaying.

waharnum:
Now that the APCP meeting is finished, I'm hoping this can be looked at again with an eye towards merging. @amb26?

the-t-in-rtf:
Would @michelled or another committer involved here be willing to help review Alan's last changes?

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
```diff
@@ -0,0 +1,49 @@
...
+The list below summarizes each section of the tutorial and the core concept covered in it. Each section introduces a new concept and connects it with previous ones.
+
+1. [Components](DeveloperIntroductionToInfusionFramework-Components.html) for organizing programs out of related pieces of data, functionality and interaction.
+
+2. [Invokers](DeveloperIntroductionToInfusionFramework-Invokers.html) to allow components to provide a consistent structure for collaboration with other components and enable overriding function behavior.
+
+3. [Events and Inversion of Control](DeveloperIntroductionToInfusionFramework-EventsAndInversionOfControl.html) to allow our programs to be built out of loosely coupled parts, and to manage sequencing and lifecycle throughout that loosely coupled structure.
+
+4. [Model Components](DeveloperIntroductionToInfusionFramework-ModelsAndModelComponents.html) that can not only track mutable data, state or content, but coordinate and synchronize their data with other model components and fire events when their state changes. [Model Listeners](DeveloperIntroductionToInfusionFramework-ListeningToModelChanges.html) are used to respond to changes in model state.
```
michelled: I think this would be clearer stated without the 'not only' and replacing 'but' with 'and'.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
```diff
@@ -0,0 +1,73 @@
...
+`fluidTutorial.helloWorld`; using namespaces helps us avoid
+potential collision of component or function names when integrating with
+other code.
+</div>
+
+## Grade Inheritance in Infusion
+<div class="infusion-docs-note">While we won't get into grade inheritance until later (combining existing grade definitions by supplying multiple items in the `gradeNames` array), we'll quickly note two important rules:
+
+<ol>
+<li>The order of precedence when grades configurations have keys with the same name in the same place is that the rightmost grade will take precedence.</li>
```
michelled: I think it should be 'grade configurations' instead of 'grades configurations'

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ModelsAndModelComponents.md
```diff
@@ -0,0 +1,106 @@
...
+layout: default
+category: Tutorials
+---
+
+Mutable data is typically expected to be stored on a component's [model](../FrameworkConcepts.md#model-objects). Component models in Infusion are altered through the [ChangeApplier](../ChangeApplier.md) functionality, which works to:
+
+* coordinate model state between different components
+* allow changes to a component's model to be listened for and responded to as though they are events
+
+What kinds of data is suitable for storing on a component model? The advice from the framework explanation of [model objects](../FrameworkConcepts.md#model-objects) is:
```
michelled: "What kinds of data are"...

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-SubcomponentsAndModelRelaying.md
```diff
@@ -0,0 +1,89 @@
...
+title: Developer Introduction to the Infusion Framework - Subcomponents and Model Relaying
+layout: default
+category: Tutorials
+---
+
+A component can include other components within its configuration; these are referred to in Infusion as subcomponents.
+
+It's also common to want related components to coordinate state through their models; we can handle this through the [model relay](../ModelRelay.md) features.
+
+The evolving "Hello, World!" component below splits out the two "hello" functions (console and web page) into separate subcomponents, and synchronizes the "hello" message through model relay. We'll also update the web page display to use a model listener as to respond to model changes.
```
michelled: The end of the last sentence should be "a model listener to respond to model changes."

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OverridingInvokersAndRefactoring.md
```diff
@@ -0,0 +1,112 @@
+---
+title: Developer Introduction to the Infusion Framework - Overriding Invokers and Refactoring
+layout: default
+category: Tutorials
+---
+
+With the console and display functionality extracted as a separate components, it's easier to see that large blocks of their configuration are similar:
```
michelled: ..."extracted as separate components"

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-TransformingModelRelays.md
```diff
@@ -0,0 +1,157 @@
...
+helloWorld = fluidTutorial.helloWorld({
+  model: {
+      message: "Hello, world!"
+    }
+});
+```
+
+A main goal of Infusion is to enable us where possible to express our applications as [declarative configurations](../FrameworkConcepts.html#declarative-configuration), reducing the overhead for other developers (or ourselves, after we have stepped away from the code for some time...) to understand the flow of program data.
+
+In this case, transforming model relays reduce the amount of code we need to write to express a common need in applications: responding to a data change in one part of the program in another part in a way more complex than simple synchronization of data.
```
michelled: This sentence is a little awkward, can you rework it?

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ConcludingRemarks.md
```diff
@@ -0,0 +1,32 @@
...
+* Infusion has strong supports for [multimodal implementations](DeveloperIntroductionToInfusionFramework-ExtendingDesignsWithExistingComponents.md) that allow programs to adapt themselves to new contexts for input and presentation.
+* We can use [transformations as part of model relays](DeveloperIntroductionToInfusionFramework-TransformingModelRelays.md) to express more complex coordination of data between different parts of a program.
+
+## Where to Go Next?
+
+* Get [your own copy of Infusion from our GitHub repository](https://github.com/fluid-project/infusion).
+* Read about the [framework concepts](/FrameworkConcepts.md) to learn more about Infusion's design goals and overall philosophy.
+* Read [further documentation on Infusion](/), including many advanced topics not covered here such as [using Infusion in node.js](/NodeAPI.md), [model transformation](/ModelTransformationAPI.md) and [unit testing](/jqUnit.html).
+* [Get involved](https://wiki.fluidproject.org/display/fluid/Get+Involved) in the [Fluid Project](https://wiki.fluidproject.org/pages/viewpage.action?pageId=3900010), the open-source software community that supports, works with and evolves the Infusion framework.
```
michelled: Should the fluid project link go to http://fluidproject.org/

michelled:
This is excellent! I found a few minor things that I put comments about. Once you fix these we should merge this in.

waharnum:
@michelled Thank you very much for the edits and comments - I think I've addressed everything now.

gtirloni:
Can we merge this?

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
```diff
@@ -0,0 +1,73 @@
...
+
+ * [`fluid.component`](../ComponentConfigurationOptions.md#options-supported-by-all-components-grades), the simplest type with support for events and public function definition (called [invokers](../Invokers.md) in Infusion).
+ * [`fluid.modelComponent`](../ComponentConfigurationOptions.md#model-components), which adds support for mutable models, model events, and model state coordination and relay between different components.
+ * [`fluid.viewComponent`](../ComponentConfigurationOptions.md#view-components), which supplement model components with functionality to bind to the DOM of a web page.
+
+Infusion's approach to component-based development emphasizes declarative configuration, loose coupling and flexible hierarchies. The framework manages component lifecycle and scoping and provides many supports to allow components to work in collaboration with one another.
+
+A new Infusion component grade is defined using the `fluid.defaults` function and a very basic "Hello, World!" component might look like this:
+
+<div class="infusion-docs-note">You can check out the [Live Example of the code below on CodePen](http://codepen.io/waharnum/pen/oBYvwx?editors=1111)</div>
```
jobara: The note is missing the `<strong>Note:</strong>` at the beginning. Also the link should be written using HTML because it is in a `<div>`

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
```diff
@@ -0,0 +1,73 @@
...
+
+```
```
jobara: You can mark this as javascript so that it will get the syntax highlighting

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
```diff
@@ -0,0 +1,73 @@
...
+var helloWorld = fluidTutorial.helloWorld();
+```
+
+
+Right now this component doesn't do anything, but we will evolve its definition throughout this introduction to demonstrate further core concepts of the framework.
+
+## Namespaces in Infusion
+<div class="infusion-docs-note">
```
jobara: Similar to the previous comment about notes and using MD syntax within HTML.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
```diff
@@ -0,0 +1,73 @@
...
+    // grades have the same key in the same position
+    gradeNames: ["fluid.component"]
+});
+```
+
+After definition, instances of the component can be created by calling the component's name as a function, with the option to pass in further configuration options as an argument:
+
+```
+var helloWorld = fluidTutorial.helloWorld();
+```
```
jobara: similar to above regarding javascript syntax highlighting

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
jobara: Also, i'm not sure that the entire section should be within a note.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
```diff
@@ -0,0 +1,73 @@
...
+<div class="infusion-docs-note">While we won't get into grade inheritance until later (combining existing grade definitions by supplying multiple items in the `gradeNames` array), we'll quickly note two important rules:
+
+<ol>
+<li>The order of precedence when grade configurations have keys with the same name in the same place is that the rightmost grade will take precedence.</li>
+<li>Any combination including `fluid.viewComponent` or grades derived from it should be placed to the right of any non-`viewComponent` grade.</li>
+</ol>
+
+</div>
+
+Next: [ Invokers](DeveloperIntroductionToInfusionFramework-Invokers.html)
```
jobara: same as previous comments.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Grade Inheritance in Infusion
+<div class="infusion-docs-note">While we won't get into grade inheritance until later (combining existing grade definitions by supplying multiple items in the `gradeNames` array), we'll quickly note two important rules:
+
+<ol>
+<li>The order of precedence when grade configurations have keys with the same name in the same place is that the rightmost grade will take precedence.</li>
+<li>Any combination including `fluid.viewComponent` or grades derived from it should be placed to the right of any non-`viewComponent` grade.</li>
+</ol>
+
+</div>
```
jobara: same as previous comments.

jobara:
@waharnum I left some comments in DeveloperIntroductionToInfusionFramework-Components.md regarding Notes and mixed MD and HTML syntax. Could you also review the rest of the changed files for these?

waharnum:
Thanks @jobara - I've cleaned everything up in regards to the Notes sections and the syntax highlighting, per your comments.

site-structure.json
```diff
@@ -146,6 +146,28 @@
...
+                    { "pageName": "Subcomponents and Model Relaying", "href": "/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-SubcomponentsAndModelRelaying.html" },
+                    { "pageName": "Restructuring Components", "href": "/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-RestructuringComponents.html" },
+                    { "pageName": "Overriding Invokers and Refactoring", "href": "/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OverridingInvokersAndRefactoring.html" },
+                    { "pageName": "Extending Designs with Existing Components", "href": "/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ExtendingDesignsWithExistingComponents.html" },
+                    { "pageName": "Transforming Model Relays", "href": "/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-TransformingModelRelays.html" },
+                    { "pageName": "Concluding Remarks", "href": "/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ConcludingRemarks.html" }
+
+
+
+
```
jobara: Seems to be several extra new lines.

jobara:
@waharnum DeveloperIntroductionToInfusionFramework.html is interesting. I'm not sure we want to maintain redirect files in the to-do directory; which was just meant for placeholders and as a way for us to keep track of what needed to be moved over. Although I can understand why you might want to have one to keep links from breaking. Is there either another way to maintain the redirect, or can we live without it?

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-Components.md
```diff
@@ -0,0 +1,68 @@
...
+other code.
+
+## Grade Inheritance in Infusion
+
+While we won't get into grade inheritance until later (combining existing grade definitions by supplying multiple items in the `gradeNames` array), we'll quickly note two important rules:
+
+1. The order of precedence when grade configurations have keys with the same name in the same place is that the rightmost grade will take precedence.
+2. Any combination including `fluid.viewComponent` or grades derived from it should be placed to the right of any non-`viewComponent` grade.
+
+Next: [ Invokers](DeveloperIntroductionToInfusionFramework-Invokers.html)
```
jobara: This links should point at the MD version of the file. That way the link is preserved in the GitHub view, and it is converted by DocPad to point at the html version.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ConcludingRemarks.md
```diff
@@ -0,0 +1,32 @@
...
+layout: default
+category: Tutorials
+---
+
+We've come to the end of our introduction to the core concepts of the Infusion framework. There are many topics we haven't covered, but with these foundational concepts, we hope you go on to further explore Infusion.
+
+We'll summarize what we've done below, then suggest some next steps.
+
+<div class="infusion-docs-note"><strong>Note:</strong> If you have feedback about this tutorial, we'd be glad to hear it via our <a href="https://wiki.fluidproject.org/display/fluid/IRC+Channel">IRC channels</a>, <a href="https://wiki.fluidproject.org/display/fluid/Mailing+Lists">mailing lists</a> or a <a href="https://github.com/fluid-project/infusion-docs/issues">GitHub issue on the documentation repo</a></div>
+
```
jobara: I wonder if we'd prefer to have them file issues in JIRA instead. It is something to think about if we want to use the separate github issue tracker, but we have been keeping most things in JIRA.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ConcludingRemarks.md
```diff
@@ -0,0 +1,32 @@
...
+## Summing Up
+
+* We implement our programs by designing [components](DeveloperIntroductionToInfusionFramework-Components.html) that work together to provide the needed functionality. Components are Javascript objects adhering to certain syntax conventions that are created using the `fluid.defaults` function. While components have defaults, any of these defaults can be overriden at the time a specific instance of the component is created.
```
jobara: References to other documentation pages should all point at the MD version, same throughout.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ConcludingRemarks.md
waharnum: I vaguely recall I had some rationale based on a conversation with someone about asking to file documentation issues in the GitHub tracker, but I cannot for the life of me remember the details, or if this conversation actually occurred. I'll update it to point to the JIRA tracker instead.

waharnum:
@jobara regarding the redirect, on the docs platform we have one link currently in `SetUpYourEnvironment.md` that points to the to-do document location. I'm not sure if there might be other external ones.

I'd feel fine about eliminating the redirect and changing the link.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ConcludingRemarks.md
waharnum: Thanks for catching this issue - they should all be fixed now.

jobara:
@waharnum okay, lets remove the redirect for now. If it becomes an issue we can think of ways to address it. Regarding the link in the SetUpYourEnvironment.md, what is the difference between the getting started tutorial and the developer introduction? Does the latter expand on the former?

waharnum:
@jobara I've removed the redirect and (per our conversation) the link from the "Getting Started" tutorial.

site-structure.json
jobara: @waharnum the extra white spaces still seem to be here

waharnum:
@jobara missed that one, sorry - fixed now.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ConcludingRemarks.md
```diff
@@ -0,0 +1,32 @@
...
+* Components use [invokers](DeveloperIntroductionToInfusionFramework-Invokers.md) to expose functionality "publicly", provide a consistent API for collaboration with other components or use by other code, and enable function overriding when deriving new components from existing ones.
+* All components support [events and inversion of control](DeveloperIntroductionToInfusionFramework-EventsAndInversionOfControl.md); these allow our programs to be built up of loosely coupled parts, and to manage sequencing through components observing and responding to their own events or the events of other components.
+* Components that need to track mutable data, state or content should be [model components](DeveloperIntroductionToInfusionFramework-ModelsAndModelComponents.md); model components can coordinate and synchronize their data with other model components, fire events when their models are changed, and take other actions to store and respond to state changes. Model components can use [model listeners](DeveloperIntroductionToInfusionFramework-ListeningToModelChanges.md) to respond to changes in model state.
```
jobara: The "model listeners" link is broken.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-ConcludingRemarks.md
```diff
@@ -0,0 +1,32 @@
...
+* Infusion has strong supports for [multimodal implementations](DeveloperIntroductionToInfusionFramework-ExtendingDesignsWithExistingComponents.md) that allow programs to adapt themselves to new contexts for input and presentation.
+* We can use [transformations as part of model relays](DeveloperIntroductionToInfusionFramework-TransformingModelRelays.md) to express more complex coordination of data between different parts of a program.
+
+## Where to Go Next?
+
+* Get [your own copy of Infusion from our GitHub repository](https://github.com/fluid-project/infusion).
+* Read about the [framework concepts](/FrameworkConcepts.md) to learn more about Infusion's design goals and overall philosophy.
+* Read [further documentation on Infusion](/), including many advanced topics not covered here such as [using Infusion in node.js](/NodeAPI.md), [advanced features of model transformation](/ModelTransformationAPI.md) and [unit testing](/jqUnit.md).
```
jobara: All of the relative links in the "Where to Go Next?" section are broken. Probably because they are relative to the infusion docs and not the tutorials.

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
```diff
@@ -0,0 +1,49 @@
...
+
+<!-- _This page is still under construction. An archived version can be found on the
+[Developer Introduction to the Infusion Framework](http://wiki.fluidproject.org/display/docs/Developer+Introduction+to+Infusion+Framework)
+page of the [Infusion Documentation wiki](http://wiki.fluidproject.org/display/docs/Infusion+Documentation)._ -->
+
+# Introduction
+
+Here we will boil down some of the basic concepts of Infusion for developers who may be interested in it, but uncertain of where to start. There's a lot going on in the framework, but grasping some core ideas helps a great deal in moving forward and learning.
+
+This introduction summarizes some topics (and leaves out many others) that are gone into at length in the [full framework documentation](/), and is focused on developers trying to orient themselves to the framework for the first time.
```
jobara: "Full framework documentation" link is broken

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
```diff
@@ -0,0 +1,49 @@
...
+
+The list below summarizes each section of the tutorial and the core concept covered in it. Each section introduces a new concept and connects it with previous ones.
+
+1. [Components](DeveloperIntroductionToInfusionFramework-Components.md) for organizing programs out of related pieces of data, functionality and interaction.
+
+2. [Invokers](DeveloperIntroductionToInfusionFramework-Invokers.md) to allow components to provide a consistent structure for collaboration with other components and enable overriding function behavior.
+
+3. [Events and Inversion of Control](DeveloperIntroductionToInfusionFramework-EventsAndInversionOfControl.md) to allow our programs to be built out of loosely coupled parts, and to manage sequencing and lifecycle throughout that loosely coupled structure.
+
+4. [Model Components](DeveloperIntroductionToInfusionFramework-ModelsAndModelComponents.md) that can track mutable data, state or content, and coordinate and synchronize their data with other model components and fire events when their state changes. [Model Listeners](DeveloperIntroductionToInfusionFramework-ListeningToModelChanges.md) are used to respond to changes in model state.
```
jobara: "Model Listeners" link is broken

src/documents/tutorial-developerIntroduction/DeveloperIntroductionToInfusionFramework-OpeningRemarks.md
jobara: May only be broken in the github view.

waharnum:
@jobara I've fixed the link issues you've noted.

Regarding the "Getting Started With Infusion" (http://docs.fluidproject.org/infusion/development/tutorial-gettingStartedWithInfusion/GettingStartedWithInfusion.html) tutorial and this one, I've got the following observations:

* "Getting Started" presumes development on a local environment; the new tutorial uses Codepen to allow someone to try out Infusion without having to install anything locally
* "Getting Started" encourages the use of the Renderer; the new tutorial stays away from the topic
* I (personal opinion) think the "Getting Started" tutorial is more complex than the new one, in terms of introducing the core Infusion concepts.

What I'd propose for now is renaming "Getting Started" to something like "Getting Started with Infusion Component Design" - a lot of the material is about component design choices and figuring out which are appropriate (very valuable material, in more detail than what's in the new tutorial). We can keep the file names the same to avoid losing links from other places and the relationship between the tutorials is more clear.

Longer term (after the new Renderer, perhaps), I think we should rewrite the tutorial to be a more advanced coverage of building Infusion-based applications.

How does that sound?

jobara:
@waharnum that seems reasonable to me.

waharnum:
@jobara I've renamed the Getting Started tutorial accordingly and made a few changes to the wording of the opening to make it clearer by comparison to the Developer Introduction - have a look and let me know your thoughts?

src/documents/tutorial-gettingStartedWithInfusion/BasicComponentCreation-Components.md
```diff
@@ -161,13 +161,13 @@ tutorials.currencyConverterAuto.convert = function (exchangeRate, amount) {
 };
 ```
 
-You'll notice that in this case we have been able to avoid binding to the entire component instance in our public function, and so our standalone public function `tutorials.currencyConverterAuto.convert` 
-is of more general utility than just for building a component method. 
+You'll notice that in this case we have been able to avoid binding to the entire component instance in our public function, and so our standalone public function `tutorials.currencyConverterAuto.convert`
+is of more general utility than just for building a component method.
```
jobara: Could you fix the bug in the code example code that this refers to. The function takes in `exchangeRate` but then uses `that.options.exchangeRate` internally.

jobara:
@michelled did you finish looking over the content of this PR, is it okay to go in from the stand point. I think I've finished reviewing the structural/syntax parts now.

michelled:
Yep, I'm happy with it!

jobara:
Merged at aa4b061d02a53e6f8dfcc02989300672c45a9617

PR #122, opened 2017-06-12
FLUID-6158: Document fluid.modelPairToChanges()

simonbates:
Add documentation for fluid.modelPairToChanges().

Code pull request: https://github.com/fluid-project/infusion/pull/830


src/documents/CoreAPI.md
```diff
@@ -539,6 +539,22 @@ if you are interested.
 path segments representing this
 * `newValue: {Any}` The value to be written into `model` at the path `path`
 
+### fluid.modelPairToChanges(value, oldValue, changePathPrefix)
+
+* `value {Any}` Model value to compare
+* `oldValue {Any}` Model value to compare
+* `changePathPrefix {String|Array of String}` [optional] Path prefix to prepend to change record paths
+* Returns: `{Array of Object}` An array of change records
```
amb26: This type should really read {Array of changeRecord} and link to the section in ChangeApplierAPI which defines it.

src/documents/CoreAPI.md
```diff
@@ -539,6 +539,22 @@ if you are interested.
...
+
+Calculates the changes between the model values `value` and `oldValue`
+and returns an array of change records. The optional argument
+`changePathPrefix` is prepended to the change path of each record (this
+is useful for generating change records to be applied at a non-root path
+in a model).
```
amb26: It would be really useful to have a super-simple example here that might allow people to get a sense of what is going on - for example, a computation of the changes between {a: 1} and {b:2}

src/documents/CoreAPI.md
```diff
@@ -539,6 +539,22 @@ if you are interested.
...
+
+The returned array of change records may be used with
+`fluid.fireChanges()`.
```
amb26: This API isn't actually documented - I wonder if I could prevail on you to add some basic docs for it in ChangeApplierAPI.md at the end of the section "Programmatic style for triggering a change" and then link to it here - cheers!

src/documents/CoreAPI.md
simonbates: I think I was incorrect in referring to the output as "change records". Looking at the ChangeApplierAPI docs, I think they are changeRequest objects, because I output "path" properties rather than "changePath" properties. Does this seem right?

I've updated the docs to read "changeRequest" and added a link to the docs for it.

src/documents/CoreAPI.md
simonbates: Great suggestion. Done.

src/documents/CoreAPI.md
amb26: Yes, this is brilliant, thanks!

src/documents/CoreAPI.md
amb26: Great example, very clear

src/documents/CoreAPI.md
amb26: I think this comment got displaced through collateral changes - do you think it seems doable?

simonbates: Yes, absolutely. I will add docs for fluid.fireChanges.

simonbates: Done.

PR #129, opened 2017-10-02
FLUID-6205: Replace "marked" with "markdown-it"...

the-t-in-rtf:
See [FLUID-6205](https://issues.fluidproject.org/browse/FLUID-6205) for details.


jobara:
@jhung would you like to review this?

the-t-in-rtf:
In reviewing this further I see that the heading links aren't working at the moment.  I will work on fixing that and comment when this is ready for review.

the-t-in-rtf:
I found a plugin that generates headers in the same way that GitHub does, which is helpful.  However, in reviewing further I discovered that at least some of our existing section links do not work on GitHub.  I wrote this up as [FLUID-6206](https://issues.fluidproject.org/browse/FLUID-6206).

My proposal is that the header links should match what GitHub generates, and that it's time to check links in the docs as outlined in [FLUID-5176](https://issues.fluidproject.org/browse/FLUID-5176).

the-t-in-rtf:
I have been working on a link checker for both placeholder links (page#id) and links between pages.  I need to harden this further, but in running the tool I found 81 internal links in the docs that were broken.  For example, the link to `fluid.flatten` on this page matches the ids our docpad site generates, but no longer matches what GitHub generates:

https://github.com/fluid-project/infusion-docs/blob/master/src/documents/IoCTestingFramework.md#modules-tests-and-sequence

I will proceed tomorrow to fix the broken links and update the pull.

amb26:
Yes, it has been observed that the docpad anchor format is different from the github one in cases where the ids contain non-word characters. In the case there is a conflict we should make sure that we observe the docpad one since this is the format in which the vast majority of users will browse the docs. I imagine it will not be possible to satisfy both formats at the same time.

jhung:
@the-t-in-rtf I noticed that the IDs are not being generated for headers, thus breaking any internal links.

For example: on `ModelTransformationAPI.html` the first link "model" goes to `FrameworkConcepts.html#model-objects`. But if you inspect the header for "Model Objects", the id "model-objects" is missing.

Also, anchors for headers are missing HREF values.

This is what the in-production markup looks like:
```
<h3 id="literal-value-fluid-transforms-literalvalue-">
   Literal Value (fluid.transforms.literalValue)
   <a class="infusion-docs-anchor" aria-hidden="true" href="#literal-value-fluid-transforms-literalvalue-">
      <span class="octicon octicon-link" style="visibility: hidden;"></span>
   </a>
</h3>
```

This is what is generated by markdown-it:
```
<h3>
   <a class="infusion-docs-anchor" aria-hidden="true" href="#">
      <span class="octicon octicon-link" style="visibility: hidden;"></span>
   </a>
   Literal Value (fluid.transforms.literalValue)
</h3>
```

the-t-in-rtf:
@jhung, I haven't committed the work that generates the IDs, as the IDs are different, and match what GitHub itself generates.  I am working to commit those changes and updated links once I confirm they're all correct.

Do please look at the link to GitHub, that's master, and not my work, and you can see that the current IDs generated by marked no longer match what GitHub itself generates.

the-t-in-rtf:
Sorry for the earlier confusion @jhung, it was not awesome to describe the problem without giving you a way to see it yourself.  I have committed (separately):

1. The configuration changes to generate the header ids.
2. The new link checking harness (which I will add tests for and clean up shortly).
3. Updated links detected by the link checking harness.

I refined the link checking late yesterday, at the high point there were 141 broken links.  Many of these were differences between the old plugin and GitHub/markdown-it.  A few were outright broken, I found typos, inappropriate slashes in one path, a broken link from the tutorials to the main docs, and quite a few other variations.

Here's an example that would have been hard to pick up on before, a link from the tutorial to the docs for fluid.transforms.linearScale was broken [in the "currency converter" example](https://github.com/fluid-project/infusion-docs/blob/master/src/documents/tutorial-gettingStartedWithInfusion/RendererComponents.md?utf8=✓#example-currency-converter).  It just needed a leading `../`.

For the conversion to using markdown-it, my idea was to look at the GitHub links and replace with those, to confirm that we are indeed creating links that will work in both.  This worked in the vast majority of cases.  However, there is now a single header that is different between GitHub and markdown-it.  I will update the link to match what's needed for the documentation site and either fix or document the issue further.

So, we're down from 141 links that didn't work when viewing the content on GitHub, to 1, and I'd estimate a few dozen broken links within the documentation site itself are fixed as part of this.

the-t-in-rtf:
I added tests for the link checker today and discovered that it wasn't properly checking page-to-page links within the docpad instance, including those to headings.  I fixed this and fixed all of the broken links revealed.

Running `npm test` in this package will now:

1. (re)generate the docpad static content.
2. Check all links within the docpad instance, including links to headings.
3. Check the link checker itself.

Anyway, @jhung, this should be ready for review.  I will keep looking at the one link that only works in the docpad instance, and either fix or (more likely) file a bug report.

the-t-in-rtf:
As an example of the stuff I found today, here's [a broken link to the nonexistent "togo" subdirectory](https://github.com/fluid-project/infusion-docs/blame/98fa4e0a222108d0acd337118ba645083d43cfee/src/documents/ContextAwareness.md#L158) that this pull now fixes.

the-t-in-rtf:
I have filed [FLUID-6207](https://issues.fluidproject.org/browse/FLUID-6207) for the single header id that would not match between GitHub and the docs site if this pull were merged.  As @amb26 suggested, perfect compatibility is more of an ideal, we can live with one known issue like this as long as it is documented and we have an informal agreement to revisit if it comes up more often.

the-t-in-rtf:
@jhung and I chatted, I wanted to add this to the discussion here:

> **jhung: Hi Tony. I'm getting an error when trying to run npm test. It seems to be a path problem.** 
> **jhung: ASSERTION FAILED: Module infusion-docs has not been loaded and could not be loaded from caller's path /Users/jhung/Work/development/infusion-docs/tests/js**
> the-t-in-rtf: That's odd
> the-t-in-rtf: infusion-docs should be automatically picked up by fluid.require
> the-t-in-rtf: what platform?  OS X?
> **jhung: yep OS X**
> the-t-in-rtf: It should have picked the module up from infusion-docs/index.js
> the-t-in-rtf: does that exist for you?
> the-t-in-rtf: If so, what does it look like?
> **jhung: let me check...**
> **jhung: "use strict";**
> **jhung: var fluid = require("infusion");**
> **jhung: fluid.module.register("infusion-docs", __dirname, require);**
> the-t-in-rtf: yup, that's right.
> the-t-in-rtf: did you clear out the node_modules directory when you pulled the latest changes?
> the-t-in-rtf: If you're still stuck after that, show it to Antranig, this is firmly his territory.
> the-t-in-rtf: fluid.require should "just work"
> **jhung: yes. I cleared that. I'm trying again. I updated my version of npm. Not sure if that will make a difference.**
> the-t-in-rtf: I'm on npm 3.10.10 and node 6.10.3
> **jhung: hmm. I'm on npm 3.10.10 and node 6.11.4**
> **jhung: Would it help if I emailed you the stack trace from my console?**
> the-t-in-rtf: Not sure, as I can't reproduce anything like it locally.
> the-t-in-rtf: I think it would mean more to Antranig in the abstract.
> **jhung: hmmm ok. I'll see if I can have Justin give this a try.**	
> the-t-in-rtf: Sounds good, seems very odd to me.

The new `index.js` stub should have taken care of this, I issued `fluid.require("%infusion-docs")` from each of the test files.  @amb26, we might need your help here, as I can't reproduce this myself.

For now, I will leave a note in the code and replace both calls to `fluid.require("%infusion-docs")` with `require("../../")` until @amb26 can review.  @jhung, that should allow you to continue the review.

amb26:
Hi there @jhung - could you please back up to the problematic commit and paste a full transcript of the output you get when you run npm test? Thanks

jhung:
@amb26 here's the output I get from that commit. This repeats 3 times each time I execute `npm test`.

```
Infusion at path /Users/jhung/Work/development/node_modules/infusion is at top level 
09:14:54.945:  Registering module infusion from path /Users/jhung/Work/development/node_modules/infusion
09:14:54.946:  Registering module infusion from path /Users/jhung/Work/development/node_modules/infusion
09:14:55.079:  ASSERTION FAILED: Module infusion-docs has not been loaded and could not be loaded from caller's path /Users/jhung/Work/development/infusion-docs/tests/js
09:14:55.079:  Current activity: 
09:14:55.080:  FATAL ERROR: Uncaught exception: Assertion failure - check console for more details: Module infusion-docs has not been loaded and could not be loaded from caller's path /Users/jhung/Work/development/infusion-docs/tests/js
Error: Assertion failure - check console for more details: Module infusion-docs has not been loaded and could not be loaded from caller's path /Users/jhung/Work/development/infusion-docs/tests/js
    at new Error (native)
    at Error.fluid.FluidError (/Users/jhung/Work/development/node_modules/infusion/src/framework/core/js/Fluid.js:173:26)
    at Object.fluid.builtinFail (/Users/jhung/Work/development/node_modules/infusion/src/framework/core/js/Fluid.js:197:15)
    at jqUnit.nodeFailureHandler (/Users/jhung/Work/development/infusion-docs/node_modules/node-jqunit/lib/jqUnit-node.js:38:11)
    at Object.fire (/Users/jhung/Work/development/node_modules/infusion/src/framework/core/js/Fluid.js:1493:40)
    at Object.fluid.fail (/Users/jhung/Work/development/node_modules/infusion/src/framework/core/js/Fluid.js:212:32)
    at Object.fluid.require (/Users/jhung/Work/development/node_modules/infusion/src/module/module.js:136:23)
    at Object.<anonymous> (/Users/jhung/Work/development/infusion-docs/tests/js/test-docpad-links.js:15:7)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/jhung/Work/development/infusion-docs/tests/all-tests.js:2:1)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:383:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:496:3
```

amb26:
Yes, the issue is that the stray copy of infusion that you have in your filespace is being picked up rather than the one inside infusion_docs - this is because it is in a directory named node_modules - this is a practice which will lead to all kinds of problems so I recommend you move this copy away and/or rename it

the-t-in-rtf:
Thanks for working to figure that out, @jhung and @amb26.  I have reverted the previous workaround, as I prefer to use `fluid.require` wherever possible.

jhung:
@amb26 Thanks for pointing that out, I didn't even realize that there was a stray `node_modules` directory one level up (it's been there since November 2016). Removing it fixed this problem. Will continue looking at the rest of this PR now.

package.json
```diff
@@ -1,19 +1,30 @@
...
-    },
-    "dependencies": {
-        "URIjs": "^1.14.1",
-        "docpad-plugin-ghpages": "~2.4.4",
-        "docpad-plugin-handlebars": "~2.3.0",
-        "docpad-plugin-highlightjs": "~2.2.2",
-        "docpad-plugin-marked": "~2.2.1",
-        "fs-extra": "^0.18.0"
-    }
+  "name": "infusion-docs",
```
amb26: Is there a way you could get your editor to revert to 4-spacing in JSON files? Unfortunately our linter isn't good enough to pick this up, but it is what we use in other files and it tends to make unnecessary diffs when, e.g. package.json files get edited

package.json
```diff
@@ -1,19 +1,30 @@
...
+  "license": "CC-BY-3.0",
+  "repository": "https://github.com/fluid-project/infusion-docs",
+  "scripts": {
+    "docpad": "docpad run --env static",
+    "deploy": "docpad deploy-ghpages --env static",
+    "pretest": "docpad generate --env static",
+    "test": "node tests/all-tests.js"
+  },
+  "dependencies": {
+    "URIjs": "^1.14.1",
```
amb26: Given this diff has been generated for the time being, we should take the opportunity to adjust these to our project standard semver usage of only binding to exact versions

src/documents/ContextAwareness.md
```diff
@@ -155,7 +155,7 @@ The `checkRecord` structure which is used in the first row of the table above is
 
 ### Example `contextAwareness` record
 
-The most adaptible component in the framework is currently the [Uploader](togo/Uploader.md) which currently can respond to three "dimensions" of adaptation. Two of these,
```
amb26: Hilarious

src/documents/FrameworkConcepts.md
```diff
@@ -30,7 +30,7 @@ It is possible to replace one object tree with another, but still to maintain a
 This is particularly important in web applications where data claiming to be "your data" can suddenly arrive from anywhere (a JSON feed, some persistence,
 a particularly aggressive version management system, etc). However it got here, you know it is your data because it is at the right path.
 
-EL paths within Infusion can be evaluated (dereferenced with respect to a particular model) by the framework utilities [`fluid.get()`](CoreAPI.md#fluid-get-model-path-) and [`fluid.set()`](CoreAPI.md#fluid-set-model-path-newvalue-),
+EL paths within Infusion can be evaluated (dereferenced with respect to a particular model) by the framework utilities [`fluid.get()`](CoreAPI.md#fluidgetmodel-path) and [`fluid.set()`](CoreAPI.md#fluidsetmodel-path-newvalue),
```
amb26: I do like the new style of not requiring trailing "-" junk, but I can't quite yet spot the regularity behind the rule of which characters get converted to "-" and which don't

package.json
the-t-in-rtf: Absolutely, I only avoided it because I thought there might be some reason for it.

the-t-in-rtf: Done.

package.json
the-t-in-rtf: Yes, there is, and I updated the setting.  There may be a few more unnecessary diffs as various pieces of work cross to the other side.

src/documents/FrameworkConcepts.md
the-t-in-rtf: The plugin we use ([`markdown-it-anchor`]()) uses the library [`string.js`](https://github.com/jprichardson/string.js), which provides a `slugify` function:

https://github.com/jprichardson/string.js/blob/master/lib/string.js#L417

That function calls a `dasherize` function from the same package:

https://github.com/jprichardson/string.js/blob/master/lib/string.js#L128

As I suspected from the variations I'd seen, it replaces one *or more* whitespace characters with a single dash.  Their algorithm seems to match what GitHub does in the 150 or so cases I tested, with the exception of [the one pattern I found](https://issues.fluidproject.org/browse/FLUID-6207).  We could point a link checker at GitHub to confirm if there are other variations, but unless we can find details about the logic GitHub itself uses (I couldn't in 15-20 minutes of searching), we would be making just as much of a guess ourselves in picking a replacement or writing our own plugin.

On a side note, in searching I discovered that GitLab (which a few of us have used for its CI integration) [helpfully publishes their rules](https://docs.gitlab.com/ee/user/markdown.html#header-ids-and-links), but insanely, even visually reviewing them it's obvious they don't match GitHub at all.

the-t-in-rtf: I read through the GfM spec in some depth, the closest I could find is in [the section on links](https://github.github.com/gfm/#links):

> One label matches another just in case their normalized forms are equal. To normalize a label, strip off the opening and closing brackets, perform the Unicode case fold, strip leading and trailing whitespace and collapse consecutive internal whitespace to a single space. If there are multiple matching reference link definitions, the one that comes first in the document is used. (It is desirable in such cases to emit a warning.)

This sounds very close to what `string.js` does, with the additional step of replacing the remaining spaces with dashes.  However, obviously there is something in `fluid.log([logLevel, arg1, ... argn])` that is handled differently by the two, that causes `, ... ` to end up as a single dash with `string.js` and as a double-dash on GitHub (see [FLUID-6207](https://issues.fluidproject.org/browse/FLUID-6207)).

amb26:
Thanks, this work looks great - I was able to get the link checker to work after running the "docpad generate" task as is done by the "pretest" script, but I notice that the "docpad run" task as run by the "docpad" task puts this output somewhere else, where the link checker can't find it - is there a way we could harmonise these two scripts?

Also, given we have now accumulated a significant amount of .js code within this project, could you please import our standard eslint tasks and configs as seen, e.g. in https://github.com/fluid-project/kettle/blob/master/Gruntfile.js#L15 and check that this code lints properly?
Cheers!

the-t-in-rtf:
@amb26, as we discussed in the chats, I added linting (all standalone json and javascript files passed).  I took a first pass at markdown rules, which should be reviewed.

I wanted to point out that it's also possible to [lint the javascript code found in our pages](https://github.com/ChrisWren/grunt-mdlint).  I know from my recent review that we use snippets that end in commas, shorthand like <insert stuff here> that won't pass, so there would be some work required.

Gruntfile.js
```diff
@@ -0,0 +1,49 @@
+/*
+Copyright 2013-2014 OCAD University
```
jobara: Is this Copyright correct? Was this file copied from somewhere else, or does it need to be updated/changed?

src/documents/ChangeApplierAPI.md
```diff
@@ -244,7 +245,7 @@ The values of built-in sources supported as values in `excludeSource` and `inclu
     </tbody>
 </table>
 
-##### Example featuring built-in change source filtering:
+##### Example featuring built-in change source filtering: ####
```
jobara: some of the headers have imbalanced #s. for example this one has 5 on the left and 4 on the right. I think there are a few other cases like this as well.

src/documents/ComponentConfigurationOptions.md
```diff
@@ -272,7 +281,7 @@ fluid.defaults("examples.eventedComponent", {
     <th>Notes</th>
     <td>Some special context names will be available within the subcomponent's definition block, for example <code>{source}</code> and <code>{sourcePath}</code> or <code>{arguments}</code>, derived from the material responsible for constructing the component.
     <em>This framework facility will be replaced by a more declarative equivalent in time - ask on the <a href="http://lists.idrc.ocad.ca/mailman/listinfo/fluid-work">fluid-work mailing list</a>
-    or <a href="https://wiki.fluidproject.org/display/fluid/IRC+Channel">#fluid-work IRC channel</a> if you seem to find yourself needing to use it.</em></td>
+    or <a href="https://wiki.fluidproject.org/display/fluid/IRC+Channel">#fluidwork IRC channel</a> if you seem to find yourself needing to use it.</em></td>
```
jobara: The IRC channel is actually "#fluid-work".

src/documents/IoCAPI.md
```diff
@@ -86,11 +87,11 @@ of view of component `that`.
 
 ### fluid.parseContextReference(reference) {
 
-Parse an [IoC reference](IoCReferences.md) of the form `"{myContext}.myPath1.myPath2"` into an object form. In this case, for example, the output would be 
+Parse an [IoC reference](IoCReferences.md) of the form `"{myContext}.myPath1.myPath2"` into an object form. In this case, for example, the output would be
 
 ```json
 {
-    context: "myContext", 
+    context: "myContext",
```
jobara: should we fix these malformed snippets in the process, or do that later?

src/documents/ComponentConfigurationOptions.md
amb26: Well spotted :)

src/documents/IoCAPI.md
amb26: Unless we can add in a JSON5 mode to the plugin :)
A semi-serious suggestion, I'd be willing to help out with this if it looked feasible.

src/documents/ChangeApplierAPI.md
the-t-in-rtf: So, we should discuss.  This is one of three valid styles, and we use all three styles in this package.  The only check I added was to force them to be consistent within a single file.   If we want to enforce "closed" headers, I can flip that switch and fix the outliers, but I'd estimate an hour or two of cleanup, as there were many examples of both the "open" style and the "header followed by a line of equals" style.

src/documents/ComponentConfigurationOptions.md
the-t-in-rtf: Very good, and done.  This is what I love about these shallow passes at a whole package, we get a good chance to clean up along the way.

src/documents/IoCAPI.md
the-t-in-rtf: I mentioned to Antranig that there is a confusingly-named [`grunt-mdlint` plugin](https://github.com/ChrisWren/grunt-mdlint) that claims to lint javascript code within markdown.  This would help here if we used the `javascript` type for these blocks, but would also help with other blocks that are in fact javascript.

Unfortunately, after trying the plugin this morning, it seems like it wasn't updated for the jump from npm@2 to npm@3.  It tries to make a hard-coded call to the underlying mdlint binary that does not work if mdlint is already a dependency and in your project's node_modules instead of the plugin's:

```
Error: Cannot find module '/Users/duhrer/Source/rtf/infusion-docs/node_modules/grunt-mdlint/node_modules/mdlint/bin/mdlint'
```

I will see if we have the option to plug something into mdlint via the grunt-markdownlint plugin itself.

the-t-in-rtf: There is [a plugin for ESLint](https://github.com/eslint/eslint-plugin-markdown), which seems interesting.  I'll give that a try.  I have not found anything comparable for JSON or JSON5 yet.

the-t-in-rtf: OK, I redefined the `eslint` task as a multitask.  `eslint:js` is the previous default and runs as part of the `lint` rollup command.  You can use `grunt eslint:md` to see the errors reported when attempting to validate the blocks that are flagged as `javascript`.  I kept looking, but did not find any comparable plugin to validate `json` and `json5` blocks within a markdown file.

the-t-in-rtf:
So, in this morning's review, I figured out that the standard `.eslintrc.json` hadn't been set up yet, which meant our `eslint` checks previously didn't really have any "teeth".  I fixed this and then fixed the linting errors reported.

Gruntfile.js
the-t-in-rtf: It was copied, and I changed it.  I didn't add copyright headers to the other files we created.  Is there a plugin and/or template we use in general that I can reuse here, or is it just a copy and paste for each file?

the-t-in-rtf:
I also updated all dependencies, many of which were absolutely ancient (one went from 0.8 to 4.0).

the-t-in-rtf:
I created a separate configuration file for checking "snippets", as there are some things (unused variables, "use strict", et cetera) that are less useful in that context.  I'm reviewing the remaining linting errors in `grunt eslint:md` now.

the-t-in-rtf:
OK, the eslint plugin was pretty useful, and I fixed a lot of things, including:

1. Many cases in which `fluid.defaults` was called without the closing paren and semicolon.
2. Various missing semicolons, trailing commas, indentation errors
3. Some more serious syntax errors, including:
3.1  multiple cases in which a *leading* squiggly brace was omitted when passing options to a function
3.2 multiple cases in which a transformation example had a superfluous squiggly brace inside the transform definition.
3.3 multiple cases in which there was a *missing* comma between keys in an object.
4. One hilarious use of `vary` instead of `var`.  Yes, dear, they DO vary!
5. Quite a few calls to `fluid.default` instead of `fluid.defaults`.  Thankfully `.default` is a reserved word and throws an error.
6. A number of partial options blocks that were better expressed using the `json5` language.

There are a few bits of shorthand we use.  One is using `...` (or sometimes `....`) to indicate that there is more assumed material.  The fix for this was simply to add a leading javascript comment `//`, although in many cases this meant I needed to remove the comma from the preceding line.  I also standardised on `// ...` in cases where "// more will go here" had been used previously.

Another bit of shorthand is using placeholders like `<insert something>`.  If we want to have valid code, these need to be real variable names, comments, or real variables with trailing comments.  I am open to suggestions as to which seems to make the most sense.  It's mainly a problem for keys that follow this pattern.

A final is using a code block to outline a method signature, including editorial parenthetical marks like "[optional]" next to one argument, as shown here:

https://github.com/fluid-project/infusion-docs/blame/master/src/documents/SwitchAPI.md#L12

This needs to either be broken down into multiple code examples, or treated as text.

@amb26 and @jhung, can we meet during our overlap on Friday to review together?  I think we could knock most of this out in a single working meeting, or at least review enough to agree on direction for further cleanup.

the-t-in-rtf:
I just realised that I have a conflict on Friday.  @amb26 and I often meet at 16:00 CET on Mondays, @jhung, you're welcome to join us if you're available.

src/documents/ChangeApplierAPI.md
jobara: It would be nice to have consistency across files, and thank you for making them at least consistent within the file. I'm not sure which style we want to settle on, but we may want to consider a "style guide" at some point. Regarding my comment above though, I was really just mentioning that with the closed style (assuming that the markdown above is in closed style), it'd be good to have the same number of "#" on either side. Otherwise it may be confusing about which level it actually is.

the-t-in-rtf: *facepalm*.  I was looking at what I'd removed.  You're absolutely right, it does appear unbalanced.  However, having them balance is not covered by the markdown linter, I checked with both "consistent" and "atx closed".

I went to report a feature request, but in researching for that I discovered that balancing the closing headers is in fact not part of the CommonMark spec by design:

http://spec.commonmark.org/0.28/#example-42

This might be an argument in favour of using "atx" headers (no closing), which less typing and which can be linted.  For now, I just "balanced" the closing header until we can discuss.

Speaking of which, thanks for accepting the invite to chat on Monday.  I will mention our upcoming chat at the architecture meeting, IMO a draft "style guide" would be a good next step once we've resolved the current issues and talking about some of the style concerns.  I was thinking we could make a style guide part of the documentation for the small plugin we use to standardise our rules, similar to what we've done with ESLint.

the-t-in-rtf:
Hi, team.  This morning I spiked a Grunt plugin to lint JSON(5) code-fenced blocks in our markdown.  See [GPII-2660](https://issues.gpii.net/browse/GPII-2660) and [the associated pull](https://github.com/GPII/gpii-grunt-mdjson-lint/pull/1) for details.

I got far enough to cut a dev release and use it in this package.  I fixed over half the errors, but there are a few left we should discuss:

```
Running "mdjsonlint:src" (mdjsonlint) task
>> ./src/documents/RendererComponentTreeExpanders.md(304:33):Expected 'r' instead of 'h' at line 10 column 33 of the JSON5 data. Still to read: "hat.options.strings."
>> ./src/documents/RendererComponentTreeExpanders.md(325:21):Expected 'r' instead of 'h' at line 4 column 21 of the JSON5 data. Still to read: "hat.showMediumImage("
>> ./src/documents/RendererDecorators.md(19:12):Unexpected 'a' at line 3 column 12 of the JSON5 data. Still to read: "aValue,\n    decorato"
>> ./src/documents/RendererDecorators.md(210:2):Unexpected 'd' at line 1 column 2 of the JSON5 data. Still to read: "decorators: [{\n    t"
>> ./src/documents/RendererDecorators.md(294:20):Unexpected 'c' at line 5 column 20 of the JSON5 data. Still to read: "container,\n        o"
>> Found 5 errors in 133 files...
```
All of these could be converted to "snippets", but I was wondering if we couldn't either sanitise a few of them or upgrade them to be syntactically valid javascript examples.  I'm a bit worried that all of our brokenness will end up hiding in snippets if we overuse them, I found a half dozen legitimate syntax errors.

amb26:
Issue with the current gruntfile - running "grunt lint" complains "Warning: Task "index.js" not found. Use --force to continue." - I see it is listed as a bare file in the Gruntfile task - is this something which is supported by a different version of Grunt, or something stray?

amb26:
Also we have an "mdlint" task which appears unreferenced

amb26:
The errors in RendererComponentTreeExpanders: Let's rewrite this examples as conformant JavaScript since they obviously not very declarative at all. Let's make them like the example at line 143 - that is, nested inside a properly named function which accepts (model, options) and returns a tree.

Also let us stick a large NOTE on this file indicating that its entire contents are deprecated and that anyone wanting to use it should get in touch with the team. It looks like our current style for this is at https://github.com/fluid-project/infusion-docs/blame/master/src/documents/InfusionEventSystem.md#L49 (used blame view to get line numbers)

amb26:
Renderer Decorators example 1 - this should be moved to "angle brackets" form and should most likely stay a snippet
example 2: Rewrite this as an invoker reference "{that}.hide" e.g.
example 3: "angle brackets" plus snippet

amb26:
Cheers - I see that since we started treating snippets, "npm test" is now showing a slew of errors of this form:

    Error: Highlight.js: Syntax highlight definition does not exist for language: json5

the-t-in-rtf:
I aliased "json5" to use the highlighting from "json", once I did that we started getting errors about our "snippet" type, which I aliased to "shell".  I also merged in changes from master and fixed the linting errors in the merged code.

amb26:
In the course of trying to test the link checker I find that we have another kind of regression - markup which was written within HTML tables is no longer being converted. Is there an option we can supply to "markdown-it" to restore this behaviour that we had? You can see the failed rendering by comparing the local version of the table "Full listener record form" with the one in https://docs.fluidproject.org/infusion/development/InfusionEventSystem.html

the-t-in-rtf:
Looking at [the section of CommonMark spec that covers HTML blocks](http://spec.commonmark.org/0.28/#html-block), it seems like we need to have carriage returns between blocks of HTML content and markdown.  I will see if there are checks that can help enforce this and work to update various tables.

the-t-in-rtf:
@amb26, I have a means of detecting the problem (looking for unrenderered links in the HTML output).  In reviewing the places where this is a problem, I see that the problem is mainly a result of our using markdown links within HTML.  To fix this and still have the markdown be valid, I'd basically need to break the indentation, as in the following example:

```html
        <td><code>args</code><a href="#alternatives"><sup>[&dagger;]</sup></a> [optional]</td>
        <td><code>{Object|Array}</code></td>
        <td>

arguments to be supplied to the listener function when it is called - these may contain IoC references including
references to the context `{arguments}` as described in [Listener Boiling](EventInjectionAndBoiling.md#listener-boiling). `{change}` is not currently supported.
</td>
```

The carriage returns are required by the CommonMark standard in order for the text to have the chance to be parsed as markdown.  The text to be treated as markdown must be deindented on the left to avoid being treated as a code block by virtue of the indentation.

As markdown-it is following the CommonMark standard here, there are two solutions I can see here:

1. Fix the blocks that need to be treated as markdown using carriage returns and indentation, as demonstrated above.
2. Update the links to use anchor tags.

I would suggest the latter.  We are already using a mix of anchor tags and markdown links in many places, consistently using anchor tags would result in more readable and cleanly indented code.

amb26:
My preference is for the carriage returns and indentation option. We don't just have the use of anchors, we have other markdown formatting too such as backquotes. It reduces a lot of the value of markdown if we can't use it wherever possible ...

src/static/css/infusion-docs.css
```diff
@@ -141,6 +141,11 @@ a:focus {
     font-size: 1.2rem;
 }
 
+/* override the above to avoid paragraphs in tables being resized inappropriately */
+.infusion-docs-container tr.infusion-docs-alt-a-row td p, .infusion-docs-container tr.infusion-docs-alt-b-row td p {
```
amb26: Not all appropriate table rows are matched by this form - e.g. the first row in the table "Full listener record form". 
I suggest we have a more aggressive rule of something like .infusion-docs-container tr td p

Also, unless the user remembers to add the blank line first, the p does not appear either, leading to a mishmash of sizes as you can see in the "Using an event firer procedurally" table - although hopefully fixing the inherit rule specificity will deal with this

tests/js/lib/link-checker.js
```diff
@@ -0,0 +1,166 @@
...
+};
+
+fluid.tests.docs.linkChecker.checkInternalLinks = function (that) {
+    fluid.each(that.internalLinksToCheck, function (link) {
+        var linkBeforeHash = fluid.tests.docs.linkChecker.urlBeforeHash(link);
+        var linkAfterHash  = fluid.tests.docs.linkChecker.urlAfterHash(link);
+        if (linkAfterHash && linkAfterHash.length) {
+            var pageIds = that.idsByPage[linkBeforeHash];
+            if (!pageIds || pageIds.indexOf(linkAfterHash) === -1) {
+                that.brokenInternalLinks.push(link);
```
amb26: Could we add a little extra tracking in this outgoing structure (which I guess will need to be pushed through some arguments) so that the user at least has the courtesy of knowing which file the broken link appears in? I appreciate that the location within the file will be less useful since this is generated content

tests/js/lib/link-checker.js
```diff
@@ -0,0 +1,166 @@
...
+
+            if (href && href.length) {
+                var resolvedHref = fluid.tests.docs.linkChecker.resolveSafely(pageToScan, href);
+
+                var linkBeforeHash = fluid.tests.docs.linkChecker.urlBeforeHash(resolvedHref);
+                var linkAfterHash  = fluid.tests.docs.linkChecker.urlAfterHash(resolvedHref);
+
+                // As long as this is a placeholder link somewhere within the site, hold on to it to check later.
+                if (linkAfterHash.length && resolvedHref.indexOf(that.options.baseUrl) === 0) {
+                    that.internalLinksToCheck.push(resolvedHref);
```
amb26: This is the point at which it would be helpful to push a little structure including pageToScan as well as just the URL

tests/js/lib/link-checker.js
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    that.scannedPages[pageToScan] = pageResults;
+};
+
+fluid.defaults("fluid.tests.docs.linkChecker", {
+    gradeNames: ["fluid.component"],
+    members: {
+        idsByPage:            {},
```
amb26: Docs for these members would be helpful

src/static/css/infusion-docs.css
the-t-in-rtf: I updated as you suggested and couldn't find any oddities in my review.  List content still seems a bit large to me, but I added a row to confirm that pure HTML and markdown lists are identical.

screen shot 2017-11-06 at 10 40 45

I also checked the live site, the style seems to be unchanged from before.  I can adjust that as well if you like.

tests/js/lib/link-checker.js
the-t-in-rtf: I rethought my approach and realised that we don't care whether the link is internal or external, only whether it is broken and where it appears.  I addressed this in a slightly different way than you suggest, but the link checker now outputs details about which files fail.  To demonstrate, I added a broken link to a single file, which resulted in output like:

```
11:34:55.917:  jq: FAIL: Module "Checking links within docpad-generated site..." Test name "Checking links..." - Message: There should be no broken links... - at sequence position 4 of 4
11:34:55.917:  jq: Expected: 0
11:34:55.917:  jq: Actual: 1
11:34:55.917:  jq: Source:     at Object.assertEquals (/Users/duhrer/Source/rtf/infusion-docs/node_modules/infusion/tests/test-core/jqUnit/js/jqUnit.js:137:19)
    at fluid.test.docs.docpadLinks.checkResults (/Users/duhrer/Source/rtf/infusion-docs/tests/js/test-docpad-links.js:27:12)
    at togo (/Users/duhrer/Source/rtf/infusion-docs/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:202:33)
    at wrapped (/Users/duhrer/Source/rtf/infusion-docs/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:320:26)
    at Object.fire (/Users/duhrer/Source/rtf/infusion-docs/node_modules/infusion/src/framework/core/js/Fluid.js:1534:40)
    at fluid.tests.docs.linkChecker.scanSinglePageOrFinish (/Users/duhrer/Source/rtf/infusion-docs/tests/js/lib/link-checker.js:59:40)
    at Array.invokeInvoker (/Users/duhrer/Source/rtf/infusion-docs/node_modules/infusion/src/framework/core/js/FluidIoC.js:1741:29)
    at Object.that.complete (/Users/duhrer/Source/rtf/infusion-docs/node_modules/infusion/src/framework/core/js/FluidProm .... [output suppressed at 1024 chars - for more output, increase fluid.logObjectRenderChars]
11:34:55.918:  File 'http://localhost:7777/infusion/development/SwitchAPI.html' has a broken link to 'http://localhost:7777/infusion/development/SwitchAPI.html#this-doesnt-exist'.
```

tests/js/lib/link-checker.js
the-t-in-rtf: Done.

tests/js/lib/link-checker.js
the-t-in-rtf: I handled this a different way but agree with and addressed the goal, which is to assist in actually fixing the broken link.

the-t-in-rtf:
@amb26, this should be ready for review again.

amb26:
Thanks - this is absolutely brilliant and really vital work. Merged up now - now we just need to get this job added to CI

PR #135, opened 2018-03-28
FLUID-6232: Update the UI Options documentation to include the npm install method of acquiring Infusion

shadxx7:
https://issues.fluidproject.org/browse/FLUID-6232

Updated the UI Options documentation to include the npm install method of acquiring Infusion


jobara:
@jhung would you be able to review this?

jhung:
There is a newer, and better way of building UI Options using NPM (without needing to build all of Infusion). This can be done using `grunt custom --include="uiOptions"`.

However this custom build does not include the HTML templates for UI Options - those need to be manually copied to the project afterwards.

We should use this new method of building UI Options instead.

PR #139, opened 2018-07-27
FLUID-6197: correct path to set locale on preferencesEditorLoader

BlueSlug:
- correct path to set locale on preferencesEditorLoader
- fix linting in another file

See comment here - it appears the path for setting `locale` changed recently (in https://github.com/fluid-project/infusion/commit/1a466e1f9188b23ccf8da90c249cafab953b7865#diff-4b9c9b993efab504ef58b9664e518698R67) from `{prefsEditorLoader}.settings.locale` to `{prefsEditorLoader}.settings.preferences.locale`, but the documentation was not updated.

@jobara could you have a look at this documentation change? It's very minor but we're also not sure (@waharnum writing this comment under @BlueSlug's account as we pair) that this part of the documentation is correct even with the path updated, given the other changes.

### DIFF
--- a/src/documents/LocalizationInThePreferencesFramework.md
+++ b/src/documents/LocalizationInThePreferencesFramework.md
@@ -255,7 +255,7 @@ that.msgLookup.lookup(value); // where value is either the string name or the ke
 
 ## Specifying a Localization
 
-The messageLoader is implemented using [fluid.resourceLoader](ResourceLoader.md). It takes `defaultLocale` and `locale` options for specifying which localized Message Bundle to fetch. The `locale` option specifies the localization desired. By default it is sourced from the prefsEditorLoader's settings object, `"{prefsEditorLoader}.settings.locale"`. The `defaultLocale` provides a fallback to use if the desired localization cannot be located. By default it is sourced from the prefsEditorLoader's `defaultLocale` option.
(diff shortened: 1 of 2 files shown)


amb26:
I'm not sure that this change is even correct. @jobara, @cindyli - could you comment since you were involved with the work in rehousing the preferences into their own section of the model. As I see it, the "locale" is not part of the user's preferences since it is not part of the data exposed to editing via the prefs framework's panels - and, significantly, it is not initialised with a value as part of the "starter" initialModel here. https://github.com/fluid-project/infusion/blob/1a466e1f9188b23ccf8da90c249cafab953b7865/src/framework/preferences/js/StarterGrades.js#L31

waharnum:
@amb26 / @jobara / @cindyli - for more context for the change, in the code, `locale` is sourced on the `messageLoader` subcomponent of `fluid.prefs.prefsEditorLoader` from `{prefsEditorLoader}.settings.preferences.locale` at this line https://github.com/fluid-project/infusion/blob/1a466e1f9188b23ccf8da90c249cafab953b7865/src/framework/preferences/js/PrefsEditor.js#L67

We can also see `locale` being set on `initialModel` via options distribution in the locale-based tests at https://github.com/fluid-project/infusion/blob/0d498cf2f3e08fe74927226ac9426e577c8d7682/tests/framework-tests/preferences/js/PrefsEditorTests.js#L737-L744

It does appear from the code the change was intentional and perhaps intended to allow a user-controlled preference for locale to be persisted as part of the preference set, even though there's not a currently exposed control for this in the standard set (@BlueSlug and I have created a JIRA to create one as part of our multilingual work).

We would have just asked @jobara about all this but it was late on Friday :)

BlueSlug:
After discussion with @waharnum and @jobara, we have decided that this documentation change is not necessary and the code that introduced the change should be rolled back. The locale value, to the extent that it will be stored, should not be stored with the other preferences. 

We will close this pull request and open another which will change the underlying source code to store the locale in a more appropriate location - the same as other persisted but non-preference properties like the panelIndex.

PR #142, opened 2018-09-13
FLUID-6209: Documentation for Remote Model and DataSource components

jobara:
https://issues.fluidproject.org/browse/FLUID-6209

Includes updates that are included in https://github.com/fluid-project/infusion/pull/933


jobara:
@amb26 would you be able to review this PR?

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,84 @@
...
+category: Infusion
+---
+
+The DataSource component provides the workflow for getting/setting (read/write) data from some data source. A data source may be an external database, cookie, in memory storage and etc. The specifics for accessing the data source must be provided in a concrete implementation. Additionally the payload of a request can be processed through an encoding step and other transformations during the get and set workflows.
+
+## Grades and Linkage
+
+There are two main grades `fluid.dataSource` and `fluid.dataSource.writable`. `fluid.dataSource` contains the base configuration and includes configuration for getting (reading) in from a data source. `fluid.dataSource.writable` adds in the configuration for setting (writing) to a data source. If you need to read and write to a data source, you'll need to add concrete implementations for both of these grades and make use of [grade linkage](IoCAPI.md#fluidmakegradelinkagelinkagename-inputnames-outputnames) to apply the concrete writable grade to the datasource configuration. In this way one would only need to add the `fluid.dataSource.writable` grade to gain the configuration from the concrete writable grade.
```
amb26: Should clarify "If you need to read and write to a data source " .... to make clear that these are instructions for people who need to implement a *new* kind of DataSource. Most users should be able to get what they need by supplying configuration to the existing ones. In general the docs should make a clearer distinction between the needs of users configuring existing sources and those writing new ones and/or the concrete implementation for one.

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,84 @@
...
+    gradeNames: ["fluid.dataSource.writable"]
+    // add grade specific configuration
+});
+
+fluid.makeGradeLinkage("my.dataSource.linkage", ["fluid.dataSource.writable", "my.dataSource"], "my.dataSource.writable");
+```
+
+## Encoding
+
+During the `get` (read) and `set` (write) workflows, listeners are provided as handles for encoding and transforming the payload. For example serializing and deserializing the payload on it's way to and from a server. The encoding is implemented by a supplied encoding component which includes `parse` and `render` methods. For `get` the data is run through `parse`. For `set`, the data is run through `render` when being sent to the server, and any returned values are sent through `parse`.
```
amb26: Link at this point to the docs for fluid.promise.fireTransformEvent since it will not be clear in what sense these things are really "listeners". 
Typo it's -> its

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,84 @@
...
+
+Two encoding classes are provided for use, `fluid.dataSource.encoding.JSON` (serializes/deserializes JSON data) and `fluid.dataSource.encoding.none` (provides no encoding/decoding transformation). If these do not meet the requirements necessary for the concrete dataSource, a different encoder subcomponent may be provided to the concrete dataSource grade.
```
amb26: "If these do not meet the requirements necessary" is redundant. Also the requirements are not those of the concrete dataSource but those of the user.

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,84 @@
...
+
+The `onRead`, `onWrite`, and `onWriteResponse` event sequences can be used to further modify the data, in addition to the specific encoding transformations. Because the encoding happens as part of these event sequences, other transformation may be added before or after the encoding phase. For example providing [modelTransformations](ModelTransformationAPI.md) before/after serializing/deserializing JSON.
+
+### fluid.dataSource.parseJSON(string)
+
+Deserializes a JSON string into a proper JavaScript object and returns a promise for the result. If there is an error, the promise will be rejected and contain the error message. Otherwise the promise will be resolved with the JavaScript object or `undefined` if the `string` value is falsey.
+
+* `string {String}` The JSON string to be deserialized
```
amb26: I think we have a "* Returns:" syntax for expressing return types and meaning?

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
+---
+title: RemoteModel API
+layout: default
+category: Infusion
+---
+
+`fluid.remoteModelComponent` builds on top of [`fluid.modelComponent`](ComponentGrades.md) with the purpose of providing a buffer between a local and
+remote model that are attempting to stay in sync. For example a local model is being updated by user interaction, this is sent back
+to a remote server, which in turn tries to update the local model. If additional user actions occur during the roundtrip, an
+infinite loop of updates may occur. `fluid.remoteModelComponent` solves this by restricting reading and writing to a single request at a time,
```
amb26: Should say it actually does a little more than this, in that it will make some attempt to "rebase" local changes that have occurred with respect to remote ones.

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+waiting for one request to complete before operating the next.
+
+## Supported Events
+
```
amb26: Typo Compnent

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+In addition to the standard `fluid.modelCompnent` events, `fluid.remoteModelComponent` add the following:
```
amb26: typo add -> adds

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+
+* `onFetch`,
+* `onFetchError`,
+* `afterFetch`,
+* `onWrite`,
+* `onWriteError`,
+* `afterWrite`
+
+For information on the different types of events, see [Infusion Event System](InfusionEventSystem.md).
```
amb26: We don't really have different "types" of events any more - all the non-core varieties are deprecated. Should instead say something like, for information on how events work and how to configure listeners to them, see xxxxx

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+
+<div class="infusion-docs-note">
+    <strong>Note:</strong> The <code>onFetch</code>, <code>afterFetch</code>, <code>onWrite</code> and <code>afterWrite</code> events are synthetic events that fire as a sequence. As implemented in the <code>fluid.remoteModelComponent</code>, the listeners for the event will all fire in the specified order and be passed the same arguments. However, the next listener will not trigger till the previous one has completed. That is, either returned a value or resolved/rejected a returned promise. This is useful to prevent further actions from happening until all of the listeners have completed, but unlike other event sequences, it cannot be used to pass the payload through a series of transformations.
```
amb26: Again link to fluid.promise.fireTransformEvent

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+</div>
+
+### onFetch
+
+<table>
+    <tbody>
+        <tr>
+            <th>Description</th>
+            <td>
+                Fires an event sequence the executes before <code>fetchImpl</code> is called.
```
amb26: typo the -> that

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+            </td>
+        </tr>
+        <tr>
+            <th>Type</th>
```
amb26: Need information on the argument payload

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+            <td>default</td>
+        </tr>
+        <tr>
+            <th>Parameters</th>
+            <td>
+                Any paramaters passed along to the rejected promise. Typically this is an <code>error</code> object.
```
amb26: Typo parameters. We need to give more guidance here on the expectations of the reject payload. Is it really typically an `error` object? Probably meant capital E `Error`. One convention we use a lot is for there to be an `isError: true` field if it is not an Error -

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+</table>
+
+### onWrite
+
+<table>
+    <tbody>
+        <tr>
+            <th>Description</th>
+            <td>
+                Fires an event sequence the executes before <code>writeImpl</code> is called.
```
amb26: the -> that

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+
+Two synthetic events, `onFetch` and `afterFetch`, are fired during the processing of a fetch. `onFetch` can be used to perform any necessary actions before running `fetchImpl`. `afterFetch` can be used to perform any necessary actions after running `fetchImpl` (e.g. updating the model, unblocking the queue). If promises returned from `onFetch`, `afterFetch`, or `fetchImpl` are rejected, the `onFetchError` event will be fired.
+
+![A flow diagram depicting the Fetch workflow](images/remoteModel_fetch_diagram.svg)
+
+## Write Workflow
+
+Executing the `write` invoker adds a write request and returns a promise for the result. Only one request can be in flight (processing) at a time. If a fetch or write request is in flight, the write will be queued. If a write request is already in queue, the result of that request will be passed along to the current write request. When a write request is in flight, it will trigger the `writeImpl` invoker to perform the actual request.
```
amb26: "the result of that request will be passed along to the current write request" is this really right? I assume you mean that the requests will be coalesced into a single write request sent to the server, and all representatives in the queue will receive the same response payload.

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+                    The data, from <code>that.model.local</code>, to send to the remote source.
+                </dd>
+            </td>
+        </tr>
+    </tbody>
+</table>
+
+## Fetch Workflow
+
+Executing the `fetch` invoker will add a fetch request and returns a promise for the result. Only one request can be in flight (processing) at a time. If a write request is in flight, the fetch will be queued. If a fetch request is already in queue/flight, the result of that request will be passed along to the current fetch request. When a fetch request is in flight , it will trigger the `fetchImpl` invoker to perform the actual request.
```
amb26: See comment below in the similar wording for "the result of that request" for WRite

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,316 @@
...
+
+Two synthetic events, `onFetch` and `afterFetch`, are fired during the processing of a fetch. `onFetch` can be used to perform any necessary actions before running `fetchImpl`. `afterFetch` can be used to perform any necessary actions after running `fetchImpl` (e.g. updating the model, unblocking the queue). If promises returned from `onFetch`, `afterFetch`, or `fetchImpl` are rejected, the `onFetchError` event will be fired.
+
+![A flow diagram depicting the Fetch workflow](images/remoteModel_fetch_diagram.svg)
```
amb26: Should try to say something in words explaining the "rebasing" workflow in  fluid.remoteModelComponent.updateModelFromFetch

src/documents/RemoteModelAPI.md
jobara: @amb26 it is whatever gets returned for the rejected promise reason. So in that case, I don't really know exactly what it will be because it is dependent on what happens in the onFetch promise sequence and the onFetchImpl. Any thoughts on how to better document this?

src/documents/RemoteModelAPI.md
jobara: Yes, I believe that is what I meant. I'll just make use of your text for the explanation.

jobara:
@amb26 I believe I've addressed all of your comments. Ready for more review.

jobara:
@amb26 would you be able to review this DOCs PR?

jobara:
@amb26 could you please take another look at this PR?

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,106 @@
...
+fluid.defaults("my.component", {
+    components: {
+        dataSource: {
+            type: "fluid.dataSource",
+            options: {
+                gradeNames: ["fluid.dataSource.writable"],
+                listeners: {
+                    // these would point at concrete implementations for performing the read and write operations.
+                    "onRead.impl": "my.component.doRead",
+                    "onRead.impl": "my.component.doWrite"
```
BlueSlug, line 23:
```suggestion
                    "onWrite.impl": "my.component.doWrite"
```

jobara: @BlueSlug thanks for catching this.

jobara:
@greatislander I'm getting a netlify deploy failure. I think it might be related to images. How do I reference those though properly, and can you see if there is anything else that might be causing problems?

jobara:
@amb26 The build errors have all been addressed. Ready for review.

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,135 @@
...
+                    "onWrite.impl": "my.component.doWrite"
+                }
+            }
+        }
+    }
+});
+```
+
+When implementing a new kind of dataSource, where reading and writing functionality are needed,
+[grade linkage](IoCAPI.md#fluidmakegradelinkagelinkagename-inputnames-outputnames) is required to apply the concrete
```
amb26, line 38: is required -> is used

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,135 @@
...
+writable grade to the dataSource configuration. In this way, one would only need to add the `fluid.dataSource.writable`
```
amb26, line 39: "In this way, one would need" -> "In this way, a user who has already applied the name of a particular concrete read-only dataSource only needs to add `fluid.dataSource.writable` as a further mixin in order to get a read/write implementation, regardless of what the original concrete grade was"

We should think about whether we want to backport the version of this which is in the "Infusion 4" branches which relies instead on a flag `writable: true/false` and uses ContextAwareness rather than grade linkage so that we can avoid API changes in two successive releases

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,135 @@
...
+`render` methods. For `get` the data is run through `parse`. For `set`, the data is run through `render` when being sent
+to the server, and any returned values are sent through `parse`.
+
+Two encoding classes are provided for use, `fluid.dataSource.encoding.JSON` (serializes/deserializes JSON data) and
+`fluid.dataSource.encoding.none` (provides no encoding/decoding transformation). However, other encoder subcomponents
+can be implemented and configured into the dataSource.
+
+The `onRead`, `onWrite`, and `onWriteResponse` event sequences can be used to further modify the data, in addition to
+the specific encoding transformations. Because the encoding happens as part of these event sequences, other
+transformation may be added before or after the encoding phase. For example providing
```
amb26, line 71: other transformation -> other transformations

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,135 @@
...
+1. If the user has supplied a function in place of method `options`, register this function as a success handler
+2. If the user has supplied an `onError` handler in method `options`, this is registered, otherwise we register the
+firer of the dataSource's own `onError` method.
+
+* `that {Object}` The concrete dataSource component
+* `promise {Promise}` The promise to bind the resolution and reject handlers too
+* `options {Object|Function}` (optional) Either a function to be executed on resolution, or an object containing an
+`onError` ({onError: function}) handler executed on rejection.
+
+### fluid.dataSource.defaultiseOptions(componentOptions, options, directModel, isSet)
```
amb26, line 104: This isn't an API function and shouldn't be documented

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,135 @@
...
+
+Compiles the options to be passed along to the each listener in a transform event sequence. The following properties
+will be added to the `options` passed in: `directModel`, `operation` ("set" or "get"), and `notFoundIsEmpty`.
+
+* `componentOptions {Object}` The components options.
+* `options {Object}` The base set of options to be passed along to the each listener.
+* `directModel {Object}` The direct model expressing the "coordinates" of the model
+* `isSet {Boolean}` A flag to indicate if it is for a "get" or "set" operation.
+
+### fluid.dataSource.get(that, directModel, options)
```
amb26, line 114: fluid.dataSource.get isn't itself part of the API but part of the implementation - instead we should document what the contract is as part of DataSources in general - i.e. this should be listed as something like <dataSource>get(that, directModel, options)

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,135 @@
...
+Operate the core "transforming promise workflow" of a dataSource's `get` method. The initial listener provides the
+initial payload; which then proceeds through the transform chain to arrive at the final payload. A promise is returned
+with the final resolved payload.
+
+* `that {Component}` The dataSource itself
+* `directModel {Object}` The direct model expressing the "coordinates" of the model to be fetched
+* `options {Object}` A structure of options configuring the action of this get request - many of these will be specific
+to the particular concrete DataSource
+
+### fluid.dataSource.set(that, directModel, model, options)
```
amb26, line 125: As above for fluid.dataSource.get - this in itself isn't an API function although it does basically form the contract of <dataSource>.set

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,364 @@
...
+* `onFetch`,
+* `onFetchError`,
+* `afterFetch`,
+* `onWrite`,
+* `onWriteError`,
+* `afterWrite`
+
+For information on how events work and how to configure listeners to them, see [Infusion Event System](InfusionEventSystem.md).
+
+<div class="infusion-docs-note">
```
amb26, line 27: "This is useful to prevent further actions from happening until all of the listeners have completed" is pretty unclear, as is the next sentence. We should say something like "The transform promise chain pattern is here only being used for sequencing of the effects implemented in the listeners and not to actually transform the payload passed in the listener arguments"

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,364 @@
...
+            <td>
+                Fires an event sequence that executes before <code>fetchImpl</code> is called.
+            </td>
+        </tr>
+        <tr>
+            <th>Type</th>
+            <td>default</td>
+        </tr>
+        <tr>
+            <th>Parameters</th>
```
amb26, line 53: "Parameters" in all these blocks should really be "arguments" since we are trying to document what the caller should provide

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,364 @@
...
+            <th>Default</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td><code>local</code></td>
+            <td>
+                Provides the buffered mapping of the portions of the components's model to the remote data that should
+                be kept in sync. An implementor must setup the <a href="ModelRelay.md">model relay(s)</a> between the
+                component's model data and what is contained in the <code>local</code> path.
```
amb26, line 198: Needn't be in the component itself - better would be "between the provider of the model data and this component's `local" path

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,364 @@
...
+
+### fetchImpl
+
+<table>
+    <tbody>
+        <tr>
+            <th>Description</th>
+            <td>
+                Must be supplied by an implementor to provide the concrete implementation for fetching data from the
+                remote source. It must return a promise.
```
amb26, line 263: "a promise yielding the fetche data"

src/documents/RemoteModelAPI.md
```diff
@@ -0,0 +1,364 @@
...
+
+Two synthetic events, `onWrite` and `afterWrite`, are fired during the processing of a write. `onWrite` can be used to
+perform any necessary actions before running `writeImpl` (e.g. performing a fetch). `afterWrite` can be used to perform
+any necessary actions after running `writeImpl` (e.g. updating the remote model, unblocking the queue, performing a
+fetch). If promises returned from `onWrite`, `afterWrite`, or `writeImpl` are rejected, the `onWriteError` event will be
+fired.
+
+By default, upon a successful write, the `afterWrite` event will trigger the `remote` model path to be updated with the
+values from the `local` model path. Because of this, the `writeImpl` must resolve/reject its promise after the write
+implementation has been handled by the remote end point, not simply after the request has been made.
```
amb26, line 362: "has been fully handled by the remote endpoint, not simply after the I/O for the request itself has concluded"

src/documents/DataSourceAPI.md
jobara: I'll update the docs with your suggestions for now. If you it won't be too much work to back port the `writable: true/false` flag, I think it would make sense to do that.

jobara:
@amb26 I've made the changes you recommended and merged from main. Ready for more review.

src/documents/DataSourceAPI.md
jobara: @amb26 now https://github.com/fluid-project/infusion/pull/1050 that has been merged, I've updated the docs to reflect the changes for [FLUID-6623](https://issues.fluidproject.org/browse/FLUID-6623).

jobara:
@amb26 ready for more review.

src/documents/DataSourceAPI.md
```diff
@@ -0,0 +1,204 @@
...
+### fluid.dataSource.stringifyJSON(obj)
+
+Serializes a JavaScript object into a JSON string.
+
+* `obj: {Object}` The JavaScript object to be serialized
+* Returns: `{String}` If the object is undefined, an empty string is returned.
+
+## Core
+
+### fluid.dataSource.registerStandardPromiseHandlers(that, promise, options)
```
amb26, line 108: This should also be axed from being documented as public API

jobara: Removed

PR #144, opened 2018-12-04
FLUID-6311: updated UIO page with info about new "defaultLocale" option

BlueSlug:


jobara:
@BlueSlug this looks good, but could you also document how to use the locale specific languages. It seems that you'd have to write "en_US" instead of "en-US" or "en-us".

BlueSlug:
@jobara nice catch! I've added a note detailing how to specify a locale with both language and territory. It turns out the important part is the underscore, while it seems letter casing is irrelevant. "EN_us" works just as well as "en_US". I've added a note about that as well, recommending to use the latter notation and to remain constant with the message bundle filename.

I also noticed some oddities with the `onPrefsEditorReady` event formatting, so I fixed that as well

src/documents/UserInterfaceOptionsAPI.md
```diff
@@ -127,8 +127,8 @@ For information on the different types of events, see [Infusion Event System](In
             <th>Description</th>
             <td>
                 This event fires when the UI Options interface has been rendered into the iframe.
-
-                <p><em><strong>Note:</strong> use <code>onReady</code> if the listener needs UI Options to be both rendered and ready to use.</em></p>
+                <br><br>
+                <em><strong>Note:</strong> use <code>onReady</code> if the listener needs UI Options to be both rendered and ready to use.</em>
```
jobara: You shouldn't need the `<br><br>`. It seems this "NOTE" is just marked up incorrectly. It should look more like. 

```html
<div class="infusion-docs-note">
    <strong>Note:</strong> Use <code>onReady</code> if the listener needs UI Options to be both rendered and ready to use.
</div>
```

src/documents/UserInterfaceOptionsAPI.md
```diff
@@ -163,6 +163,45 @@ var uio = fluid.uiOptions.prefsEditor(".myContainer", {
...
 
+### defaultLocale
+
+<table>
+    <tbody>
+        <tr>
+            <th>Description</th>
+            <td>
+                The <code>defaultLocale</code> option allows you to specify the default locale and language of the UI Options interface. The locale specified must have associated JSON message files present in the source tree of Infusion at the path <code>src/framework/preferences/messages/</code>. The following locales are currently available in the latest version of Infusion: English (CA), English (US), Farsi, French, Spanish.
+                <br><br>
```
jobara: See previous comment about "NOTE" formatting.

src/documents/UserInterfaceOptionsAPI.md
```diff
@@ -163,6 +163,45 @@ var uio = fluid.uiOptions.prefsEditor(".myContainer", {
...
 The options supported by UI Options are described below.
 
+### defaultLocale
+
+<table>
+    <tbody>
+        <tr>
+            <th>Description</th>
+            <td>
+                The <code>defaultLocale</code> option allows you to specify the default locale and language of the UI Options interface. The locale specified must have associated JSON message files present in the source tree of Infusion at the path <code>src/framework/preferences/messages/</code>. The following locales are currently available in the latest version of Infusion: English (CA), English (US), Farsi, French, Spanish.
```
jobara: I guess the path for the message bundles isn't restricted to that location. It should be configurable through the terms in the options.

jobara:
merged at f0bc9c4294ad44e1851ed242d0fa18047864608e

PR #138, opened 2018-06-18
INFRA-193 - Add Dockerfile and update dependencies

gtirloni:
Requires Docker >17.03 (for multi-stage build support)

PR #145, opened 2019-03-26
FLUID-6324: patch for 6.82.5 release of DocPad

waharnum:
This is a patch to allow us to build and deploy the documentation under a version of DocPad compatible with Node 10.

It makes the following substantive changes:
- bumps the DocPad version to 6.82.5
- adds the `serve` plugin to maintain static server support (DocPad made this a plugin rather than a built-in for the 6.8 line)
- uses my forked and built version of the `handlebars` plugin, updated for compatibility with the 6.8 line: https://github.com/waharnum/docpad-plugin-handlebars/tree/fixFor6.8-build - I've had an outstanding pull request on the fix for this plugin since last December, but it has yet to be merged (https://github.com/docpad/docpad-plugin-handlebars/pull/13)

This should allow us to build and deploy the documentation in our usual way, in the immediate term.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,11 @@ docpad run --env static
 
 Point your browser to: `http://localhost:9778/`
 
+## Deployment
+
+1. Generate the static site: `docpad generate --env static`
+2. Commit the contents of `out` directory to the `deploy` branch (we are currently doing this manually, as the plugin we relied on previously is broken); when changes are pushed or merged to the `deploy` branch of the `fluid-project/infusion-docs` repo, CI will pick up and deploy them automatically
+
 ## A Note on DocPad plugins
 
 Changes to DocPad plugins used by this package only take effect when your `node_modules` directory is up to date.  It is
--- a/README.md
+++ b/README.md
@@ -32,26 +32,6 @@ docpad run --env static
 
 Point your browser to: `http://localhost:9778/`
 
-## Deploy
-
-While GitHub Pages is not used to host [fluiproject.org](https://docs.fluidproject.org), our deployment process requires that the generated site be pushed to the `deploy` branch in the project repo. The contents of the `deploy` branch will automatically be served as the contents are changed.
-
-To generate and push to the `deploy` branch on the project repo run the following:
-
-```shell
-npm run deploy
-```
-
-This runs the command:
-
-```shell
-docpad deploy-ghpages --env static
-```
-
-_**WARNING:** Deploying will upload the site to the `deploy` branch of the `origin` remote. If you have cloned
-from the production repository and you have push access, you will actually run the docs publication
-workflow against the live production branch, regardless of whichever branch you happen to be working on._
-
 ## A Note on DocPad plugins
 
 Changes to DocPad plugins used by this package only take effect when your `node_modules` directory is up to date.  It is
--- a/docpad.js
+++ b/docpad.js
@@ -89,44 +89,44 @@ module.exports = {
             }
         }
     },
-    // events: {
-    //     generateBefore: function () {
-    //         // Remove the previous incarnation of index.html since it causes a faulty redirect
-    //         fs.removeSync("out/index.html");
-    //     },
-    //     writeAfter: function () {
-    //         // Copy the images
-    //         fs.copySync(imagesSrcDir, imagesDestDir);
-    //
-    //         // Copy the contents of the out directory to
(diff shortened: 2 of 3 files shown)


waharnum:
@simonbates, would you be able to review this for merge when you have a chance? You should now be able to check out a clean copy of my branch, run `npm install`, and be able to build and work on the Infusion docs in the usual way.

@jhung Assuming this approach works for Simon we should be able to use it on the Guide site and other places DocPad is used.

waharnum:
@simonbates It looks like the reason for commenting out / removing the `ghpages` plugin was because that plugin is also currently broken and not updated for the most recent version of DocPad compatible with Node 10. So I think the current situation is now as follows:
- this patch will produce the expected site in the `out` directory
- the contents of the `out` directory needs to be committed to the `deploy` branch of the repo by some means; previously this was handled by the `ghpages` plugin, but that plugin is no longer working.

Options for moving forward that I see:
1. Patch the `ghpages` plugin like we did for the `handlebars` one (I don't consider this preferable, given how little this plugin is actually doing)
2. Use another means of committing the `out` directory to the `deploy` branch (Grunt task, or maybe do manually for the moment)

The Docker support that was being worked on in https://github.com/fluid-project/infusion-docs/pull/138 would make the separate deploy step unnecessary by using a builder container, but it isn't in yet (due to being blocked by the errors this mean PR is meant to fix).

I'd recommend we go with the need to make a manual commit of `out` to the `deploy` branch for the moment, given how infrequently we update the docs. Longer-term, we'll work on getting the Docker support in. Let me know your thoughts.

simonbates:
I agree with your proposal to manually copy `out` to the `deploy` branch. I think the Docker work would be a better task to spend time on, rather than getting the `ghpages` plugin to work, or automating the committing to the `deploy` branch.

simonbates:
I see that we are also changing from `docpad-plugin-markit` to `docpad-plugin-marked`. I will do some testing, but I was wondering: do we anticipate changes as a result? I'm thinking of details like automatically generated heading anchors?

waharnum:
I'm uncertain as to what changes we might see - I believe the change was made because `docpad-plugin-markit` hadn't been updated in three years, but it was made by @jobara. 

I didn't see obvious issues but haven't tested extensively - automatically generated heading anchors seem to still be working.

jhung:
> I'm uncertain as to what changes we might see - I believe the change was made because `docpad-plugin-markit` hadn't been updated in three years, but it was made by @jobara.
> 
> I didn't see obvious issues but haven't tested extensively - automatically generated heading anchors seem to still be working.

Sorry I'm late to conversation @waharnum and @simonbates. There was a change from marked to mark-it a while ago as discussed in [FLUID-6205]([url]https://issues.fluidproject.org/browse/FLUID-6205(url)).

If I recall correctly, Marked is still being used on our other docpad sites. Fluid docs is the only one using Mark-it. I will need to verify that.

simonbates:
I've generated the upstream `master` docs with Node 8 and generated the docs for this branch with Node 10. And then compared the generated output using a diff program (Beyond Compare). I'm seeing some differences:

- `"."` is handled differently in generated anchors
    - `master`: `<h3 id="fluiddefaultsgradename-options">fluid.defaults(gradeName[, options])</h3>`
    - `FLUID-6324`: `<h3 id="fluid-defaults-gradename-options-">fluid.defaults(gradeName[, options])</h3>`
- Some minor typographical: `&#39;` instead of `'`
- `src/documents/UserInterfaceOptionsAPI.md` has changes on `master` not on this branch (I think it's fine to leave this -- I note it here for completeness)

The only one I've found so far that is significant is the change in anchors. This will break related links, such as (`fluid.hasGrade` in https://github.com/fluid-project/infusion-docs/blob/master/src/documents/CoreAPI.md): `([defaults](#fluiddefaultsgradename-options))`

simonbates:
I was trying to generate the `master` docs with Node 8 again today but was having trouble (error related to the `ghpages` plugin). And with Alan's help we were able to determine that the issue was that I had the latest version of Docpad installed globally (6.82.5). I was able to generate the docs with Node 8 and Docpad version 6.79.4, which can be installed with:

```
npm install -g docpad@6.79.4
```

waharnum:
As discussed in-channel and in-person, I am going to close this in favour of [this pull request](https://github.com/fluid-project/infusion-docs/pull/146) that uses Docker containers pinned to Node Version 8.

PR #150, opened 2020-01-10
FLUID-6447: Added instructions for npm and Github.

jhung:
Added additional instructions for installation coverage.

Github installation instructions installs the most recent dev version of Infusion, whereas the Zip and npm instructions will give v2.0.0 release. Does this discrepancy matter?

### DIFF
--- a/src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
+++ b/src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
@@ -16,31 +16,34 @@ UI Options does three things:
 
 ![Screen shot of the UI Options Component](../images/uio-showcase.png "Screen shot of the UI Options Component")
 
-This tutorial assumes that:
-
-* you are already familiar with HTML, Javascript and CSS,
-* you are familiar with what the UI Options preferences editor is and does, and
-* now you just want to know how to add it to your website.
+You can try the [UI Options demo here](http://build.fluidproject.org/infusion/demos/uiOptions/).
 
 ## Scenario
 
 You're putting together a website that you know will have a diverse audience. You'd like to allow your visitors to
 customize the presentation of the site to their individual needs, by enlarging the text or increasing the visual
-contrast, for example. This tutorial will show you how to add the Infusion [UI
-Options](http://build.fluidproject.org/infusion/demos/uiOptions/) component to your site.
+contrast, for example. This tutorial will show you how to add the Infusion UI Options component to your site.
 
 These are the basic steps to add UI Options to your website:
 
-1. [Download and install the Infusion library](#download-and-install-the-ui-options-library)
+1. [Download and install the UI Options library](#download-and-install-the-ui-options-library)
 2. [Prepare your page](#prepare-your-page)
     * [The Sliding Panel](#the-sliding-panel)
     * [The Table of Contents](#the-table-of-contents)
 3. [Add dependencies to the page](#add-dependencies-to-the-page)
 4. [Add the UI Options component](#add-the-ui-options-component)
 
-The rest of this tutorial will explain each of these steps in detail.
+The rest of this tutorial will explain each of these steps in detail. This tutorial also assumes that you can work with HTML, Javascript and CSS.
 
 ## Download and install the UI Options library
+
+Get Infusion using one of these methods:
+* [Zip Package](https://github.com/fluid-project/infusion/releases/download/v2.0.0/infusion-uiOptions-2.0.0-source.zip)
+* [Github source](https://github.com/fluid-project/infusion)
+* npm via `npm i infusion`
+
+### Zip Package Method
+This method will download and install the last release of Infusion (2.0.0).
 
 1. [Download the UI Options library](https://github.com/fluid-project/infusion/releases/download/v2.0.0/infusion-uiOptions-2.0.0-source.zip)
   (there is also a [minified version UI Options](https://github.com/fluid-project/infusion/releases/download/v2.0.0/infusion-uiOptions-2.0.0-minified.zip)
@@ -48,11 +51,35 @@ The rest of this tutorial will explain each of these steps in detail.
 2. Unzip the contents of the downloaded ZIP file to a location within your project. This will result in a new directory
    `infusion`.<div class="infusion-docs-note"><strong>Note:</strong> In this guide we will use the directory
    `my-project/lib/`.</div>
-3. Your `infusion` folder will include a single file containing all of the JavaScript you need
+3. Your `infusion` directory will include a single file containing all of the JavaScript you need
    (`infusion-uiOptions.js`), HTML templates, CSS files, and other components to get UI Options to work. You will later
    link to these files in your HTML files.
 4. Now that `infusion` is in your project directory, you can delete the `infusion-uiOptions-2.0.0.zip` (or similar name)
    from your download directory.
+
+### Github Source Method
+This method will download and install the most recent development version of Infusion.
+
+1. Clone the Infusion Github repository anywhere to your local filesystem: `git clone https://github.com/fluid-project/infusion`. This will create a directory `infusion` containing the Infusion source code.
+2. Within the `infusion` directory, run `npm install` to get Infusion's dependencies.
+3. Build a version of Infusion with UI Options by running `grunt custom --include="uiOptions" --name="uiOptions"`. This will create a build of Infusion in the directory `infusion/build/`. Note: [Get Grunt](https://gruntjs.com/) if you don't have it installed.
+4. Your `infusion/build` directory will include a single file containing all of the JavaScript you need
+   (`infusion-uiOptions.js`), HTML templates, CSS files, and other components to get UI Options to work. You will later
+   link to these files in your HTML files.
+5. Rename the directory from `build` to `infusion`, and move (or copy) the `build` directory to your project (i.e. `my-project/lib/infusion`).
+6. Optional: You can now safely delete the `infusion` directory created in step 1.
+
+### npm Method
+This method will download and install the last release of Infusion (2.0.0).
+
+1. From your project directory, run `npm i infusion`.
+2. Within the `node_modules/infusion/` directory in your project, run `npm install` to get Infusion's dependencies.
+3. Then build a version of Infusion with UI Options by running `grunt custom --include="uiOptions" --name="uiOptions"`. This will create a build of Infusion in the `node_modules/infusion/build/` directory. Note: [Get Grunt](https://gruntjs.com/) if you don't have it installed.
+4. Your `node_modules/infusion/build` folder will include a single file containing all of the JavaScript you need
+   (`infusion-uiOptions.js`), HTML templates, CSS files, and other components to get UI Options to work. You will later
+   link to these files in your HTML files.
+5. Rename the directory from `build` to `infusion`, and move (or copy) the `build` directory to your project (i.e. `my-project/lib/infusion`).
+6. Optional: You can now safely delete the `infusion` directory created in step 1.
 
 ## Prepare your page
 


src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,71 @@ UI Options does three things:
...
-This tutorial assumes that:
-
-* you are already familiar with HTML, Javascript and CSS,
-* you are familiar with what the UI Options preferences editor is and does, and
-* now you just want to know how to add it to your website.
+You can try the [UI Options demo here](http://build.fluidproject.org/infusion/demos/uiOptions/).
 
 ## Scenario
 
 You're putting together a website that you know will have a diverse audience. You'd like to allow your visitors to
```
jobara: I wonder if we should leave out the "diverse audience" part.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,71 @@ UI Options does three things:
...
 customize the presentation of the site to their individual needs, by enlarging the text or increasing the visual
-contrast, for example. This tutorial will show you how to add the Infusion [UI
-Options](http://build.fluidproject.org/infusion/demos/uiOptions/) component to your site.
+contrast, for example. This tutorial will show you how to add the Infusion UI Options component to your site.
 
 These are the basic steps to add UI Options to your website:
 
-1. [Download and install the Infusion library](#download-and-install-the-ui-options-library)
+1. [Download and install the UI Options library](#download-and-install-the-ui-options-library)
```
jobara: What do you think about removing the word "library" from here and other places?

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,71 @@ UI Options does three things:
...
 2. [Prepare your page](#prepare-your-page)
     * [The Sliding Panel](#the-sliding-panel)
     * [The Table of Contents](#the-table-of-contents)
 3. [Add dependencies to the page](#add-dependencies-to-the-page)
 4. [Add the UI Options component](#add-the-ui-options-component)
 
-The rest of this tutorial will explain each of these steps in detail.
+The rest of this tutorial will explain each of these steps in detail. This tutorial also assumes that you can work with HTML, Javascript and CSS.
```
jobara: Maybe instead write this as "The rest of this tutorial will explain each of these steps in detail, and assumes that you can work with HTML, JavaScript and CSS" or maybe add the assumption as a Note.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,71 @@ UI Options does three things:
...
 
 ## Download and install the UI Options library
 
+Get Infusion using one of these methods:
+* [Zip Package](https://github.com/fluid-project/infusion/releases/download/v2.0.0/infusion-uiOptions-2.0.0-source.zip)
+* [Github source](https://github.com/fluid-project/infusion)
```
jobara: Should be "GitHub"

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,71 @@ UI Options does three things:
...
 3. [Add dependencies to the page](#add-dependencies-to-the-page)
 4. [Add the UI Options component](#add-the-ui-options-component)
 
-The rest of this tutorial will explain each of these steps in detail.
+The rest of this tutorial will explain each of these steps in detail. This tutorial also assumes that you can work with HTML, Javascript and CSS.
 
 ## Download and install the UI Options library
 
+Get Infusion using one of these methods:
+* [Zip Package](https://github.com/fluid-project/infusion/releases/download/v2.0.0/infusion-uiOptions-2.0.0-source.zip)
```
jobara: Rather than pointing at a specific zip file, what do you think about pointing at the releases page on GitHub https://github.com/fluid-project/infusion/releases Then you can just mention to pick a zip package from one of the releases so we don't have to update this tutorial for each Infusion release if nothing else changed.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,71 @@ UI Options does three things:
...
+* [Github source](https://github.com/fluid-project/infusion)
+* npm via `npm i infusion`
+
+### Zip Package Method
+This method will download and install the last release of Infusion (2.0.0).
```
jobara: Maybe call it last stable release or something like that. We've been making a lot of dev releases, which don't show up here, but are used quite a bit in other projects.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,71 @@ UI Options does three things:
...
    `my-project/lib/`.</div>
-3. Your `infusion` folder will include a single file containing all of the JavaScript you need
+3. Your `infusion` directory will include a single file containing all of the JavaScript you need
    (`infusion-uiOptions.js`), HTML templates, CSS files, and other components to get UI Options to work. You will later
    link to these files in your HTML files.
 4. Now that `infusion` is in your project directory, you can delete the `infusion-uiOptions-2.0.0.zip` (or similar name)
    from your download directory.
 
+### Github Source Method
+This method will download and install the most recent development version of Infusion.
```
jobara: I guess there is something slightly confusing with our current Infusion process we have 

- "official/stable" releases that have undergone our full test process, been tagged, and versioned. 
- dev releases which are published to NPM and are more or less points in time that someone feels marks a "milestone" of code. These may or may not all work. Sometimes there is still a bug and a new dev release is made.
- Master, this is just the latest committed code. We try not to break master though, so it is usually relatively stable in terms of tests passing and obvious bugs, but as mentioned above things do creep in from time to time. 

When you say most recent development version here, it's not clear to me that the latest dev release and master may not point at the same code.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,71 @@ UI Options does three things:
...
+
+1. Clone the Infusion Github repository anywhere to your local filesystem: `git clone https://github.com/fluid-project/infusion`. This will create a directory `infusion` containing the Infusion source code.
+2. Within the `infusion` directory, run `npm install` to get Infusion's dependencies.
+3. Build a version of Infusion with UI Options by running `grunt custom --include="uiOptions" --name="uiOptions"`. This will create a build of Infusion in the directory `infusion/build/`. Note: [Get Grunt](https://gruntjs.com/) if you don't have it installed.
+4. Your `infusion/build` directory will include a single file containing all of the JavaScript you need
+   (`infusion-uiOptions.js`), HTML templates, CSS files, and other components to get UI Options to work. You will later
+   link to these files in your HTML files.
+5. Rename the directory from `build` to `infusion`, and move (or copy) the `build` directory to your project (i.e. `my-project/lib/infusion`).
+6. Optional: You can now safely delete the `infusion` directory created in step 1.
```
jobara: I wonder if it would be helpful to also provide all of those steps as a single set of bash instructions that can just be copied and pasted into the terminal.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,71 @@ UI Options does three things:
...
+
+### npm Method
+This method will download and install the last release of Infusion (2.0.0).
+
+1. From your project directory, run `npm i infusion`.
```
jobara: So yes, this will also install the last stable release, but you could also pick a version to install. In particular you may want to provide instructions for how to install a particular dev release, or the last dev release. I believe the latest dev release gets the "dev" tag

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,71 @@ UI Options does three things:
...
+2. Within the `node_modules/infusion/` directory in your project, run `npm install` to get Infusion's dependencies.
+3. Then build a version of Infusion with UI Options by running `grunt custom --include="uiOptions" --name="uiOptions"`. This will create a build of Infusion in the `node_modules/infusion/build/` directory. Note: [Get Grunt](https://gruntjs.com/) if you don't have it installed.
+4. Your `node_modules/infusion/build` folder will include a single file containing all of the JavaScript you need
+   (`infusion-uiOptions.js`), HTML templates, CSS files, and other components to get UI Options to work. You will later
+   link to these files in your HTML files.
+5. Rename the directory from `build` to `infusion`, and move (or copy) the `build` directory to your project (i.e. `my-project/lib/infusion`).
```
jobara: Don't remember what was the case for the 2.0.0 release, but now at least you don't need to do all this work. The npm package includes a dist directory that includes builds of infusion. There's some copying around you'll have to do to move the files where you want them though, but you don't need to run your own builds.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
jhung: I agree. For technical documentation, it is a bit too vague. I'll remove that in my next update.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
jhung: Yeah it's not necessary. I'll tweak that point to say:
"Download and install UI Options"

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
jhung: Yeah. This is getting confusing because depending which of the 3 methods you follow, you can possibly end up with 3 different versions of Infusion.

While this may not be an issue for a single user, people working in teams may end up with different versions depending on which approach they take.

Some of this can be addressed by documenting this, but it can still be confusing.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
jhung: The dist directory includes different builds, but no UIO specific. I suppose I can mention they can use one of these builds, but if they want one with just UIO, then we can do a custom build.

- infusion-all-no-jquery.js
- infusion-all.js
- infusion-framework-no-jquery.js
- infusion-framework.js

jobara: The dev releases include UIO specific builds. You can see from this one on unpkg ( https://unpkg.com/browse/infusion@3.0.0-dev.20190905T163833Z.b024bff87/dist/ ). So the next release (Infusion 3.0.0) would have it as well. Are we missing any other builds that we should be generating?

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
jhung: I don't think a Note is completely necessary as it isn't a strict requirement and may not need to be emphasized.

jhung:
@jobara I've updated the documentation and hopefully made it more clear the different methods of getting a stable release, and for development versions.

Can you take a look and see if this is clearer?

jobara:
@jhung could you please run the lint task. It seems that there are some issues with at least the line lengths.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -4,62 +4,100 @@ layout: default
 category: Tutorials
 ---
 
-The **User Interface Options (UI Options)** component allows users to transform the presentation of the user interface
-and content resources so that they are personalized to the individual user's needs.
+The **User Interface Options (UI Options)** component allows users to transform the presentation of the user interface and content resources so that they are personalized to the individual user's needs.
```
jobara: should we change all these places that say "needs" to "needs and preferences"?

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -4,62 +4,100 @@ layout: default
...
-2. Unzip the contents of the downloaded ZIP file to a location within your project. This will result in a new directory
-   `infusion`.<div class="infusion-docs-note"><strong>Note:</strong> In this guide we will use the directory
-   `my-project/lib/`.</div>
-3. Your `infusion` folder will include a single file containing all of the JavaScript you need
+Get the latest Infusion release:
+* [Zip package of an Infusion release](https://github.com/fluid-project/infusion/releases)
+* Using npm: `npm i infusion`
+
+If you prefer trying the latest development source code, you can visit [Infusion on GitHub](https://github.com/fluid-project/infusion).
+
```
jobara: I'm not sure this summary is needed as you go into detail about all of these below. If you want the summary maybe make them links to the related sections instead.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -4,62 +4,100 @@ layout: default
...
+3. Then build a version of Infusion with UI Options by running `grunt custom --include="uiOptions" --name="uiOptions"`. This will create a build of Infusion in the `node_modules/infusion/build/` directory. Note: [Get Grunt](https://gruntjs.com/) if you don't have it installed.
+4. Your `node_modules/infusion/build` folder will include a single file containing all of the JavaScript you need
    (`infusion-uiOptions.js`), HTML templates, CSS files, and other components to get UI Options to work. You will later
    link to these files in your HTML files.
-4. Now that `infusion` is in your project directory, you can delete the `infusion-uiOptions-2.0.0.zip` (or similar name)
-   from your download directory.
+5. Rename the directory from `build` to `infusion`, and move (or copy) the `build` directory to your project (i.e. `my-project/lib/infusion`).
+6. Optional: You can now safely delete the `infusion` directory created in step 1.
+
+Summary of commands:
```
jobara: All the "Summary of commands:" should probably be sub headings.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
jobara: In reading the documentation and the markup again, I don't think we should be recommending that users use `npm install` to then generate a custom build using grunt. I think we should expect that if someone is using npm to source Infusion that they are already using npm for their own project. In this case they should be using `npm install infusion --save` or `npm install infusion --save-dev` and performing some copy steps, either manual or automated through grunt or other means. 

If they want to do a custom build they should pull down the source code, either by git clone or getting a GitHub generated zip file of the repo.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -4,62 +4,100 @@ layout: default
...
+This method will download and install the current version of Infusion in development.
+
+<div class="infusion-docs-note">
+<strong>Note:</strong>
+Use the latest development source if you want to try out the latest changes and features. We recommend using a stable release for production situations by using one of the other methods above.
+</div>
+
+1. Clone the Infusion GitHub repository anywhere to your local filesystem: `git clone https://github.com/fluid-project/infusion`. This will create a directory `infusion` containing the Infusion source code.
+2. Within the `infusion` directory, run `npm install` to get Infusion's dependencies.
+3. Build a version of Infusion with UI Options by running `grunt custom --include="uiOptions" --name="uiOptions"`. This will create a build of Infusion in the directory `infusion/build/`. Note: [Get Grunt](https://gruntjs.com/) if you don't have it installed.
```
jobara: Instead of having notes here, I think it would make sense to have a pre-requisite or dependencies section at the top of each of these sections. Then someone coming to them knows what's expected for them to have installed etc. before embarking on an installation type.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -4,62 +4,100 @@ layout: default
...
+mv build/ my-project/lib/infusion
+```
+
+### Infusion Latest Development Source
+This method will download and install the current version of Infusion in development.
+
+<div class="infusion-docs-note">
+<strong>Note:</strong>
+Use the latest development source if you want to try out the latest changes and features. We recommend using a stable release for production situations by using one of the other methods above.
+</div>
```
jobara: I'm not sure we should put the recommendation here, because that isn't exactly true. However, you may want to warn integrators that the documentation may not be completely accurate when using the latest development code or npm dev release.

jhung:
> 
> 
> @jhung could you please run the lint task. It seems that there are some issues with at least the line lengths.

@jobara Is there a reason for this requirement? Hand-wrapping text (especially paragraph text) is hard to maintain.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
jhung: I just checked `package.json` and grunt is a dev dependency... so if the Note even applicable? Grunt would have been installed with `npm install` correct?

jobara:
> > @jhung could you please run the lint task. It seems that there are some issues with at least the line lengths.
> 
> @jobara Is there a reason for this requirement? Hand-wrapping text (especially paragraph text) is hard to maintain.

It's part of the markdown linting. I believe it's to make the files more readable, in particular what there is no soft wrapping available. This sometimes happens in some IDEs based on user settings as well as viewing the source online. I think it's something we can revisit, and probably something that is questionable if tabs are used instead of spaces, for indentation.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,106 @@ UI Options does three things:
 
 ![Screen shot of the UI Options Component](../images/uio-showcase.png "Screen shot of the UI Options Component")
 
-This tutorial assumes that:
+You can try the [UI Options demo here](http://build.fluidproject.org/infusion/demos/uiOptions/).
```
jobara, line 19: Let's remove the word "here". 
```suggestion
You can try the [UI Options demo](http://build.fluidproject.org/infusion/demos/uiOptions/).
```

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,106 @@ UI Options does three things:
...
 
-* you are already familiar with HTML, Javascript and CSS,
-* you are familiar with what the UI Options preferences editor is and does, and
-* now you just want to know how to add it to your website.
+UI Options is also available as a [Wordpress plugin](https://github.com/fluid-project/uio-wordpress-plugin) based on the
+latest stable release of Infusion, and as a [Drupal plugin](https://www.drupal.org/project/fluidui) (maintained by Open
+  Concept).
```
jobara, line 23: We should link to OpenConcept 

```suggestion
UI Options is also available as a [Wordpress plugin](https://github.com/fluid-project/uio-wordpress-plugin) based on the
latest stable release of Infusion, and as a [Drupal plugin](https://www.drupal.org/project/fluidui) (maintained by [OpenConcept](https://openconcept.ca)).
```

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -16,44 +16,106 @@ UI Options does three things:
...
+6. Optional: You can now safely delete the `infusion` directory created in step 1.
+
+#### Summary of commands:
+
+```bash
+npm i infusion
+cd node_modules/infusion/
+grunt custom --include="uiOptions" --name="uiOptions"
+mv build/ my-project/lib/infusion
+```
```
jobara, line 82: I'm not sure this is correct. Maybe it's because we've moved the "latest" tag to point at a dev release.

Anyways, I think the NPM method should talk about installing via NPM, maybe assuming they are using a package.json file themselves. And it should not the parts that will need to be exposed or copied over from the "node modules" directory. For example sourcing the compiled JS file and Assets from "Dist" and any other material like templates and message bundles from source.

src/documents/tutorial-userInterfaceOptions/UserInterfaceOptions.md
```diff
@@ -71,7 +133,9 @@ Insert the following markup at the beginning of your `<body>` tag:
         <!-- This div is for the sliding panel that shows and hides the Preference Editor controls -->
         <div class="fl-panelBar">
             <span class="fl-prefsEditor-buttons">
-                <button id="reset" class="flc-prefsEditor-reset fl-prefsEditor-reset"><span class="fl-icon-undo"></span> Reset</button>
+                <button id="reset" class="flc-prefsEditor-reset fl-prefsEditor-reset">
+                    <span class="fl-icon-undo"></span>Reset
+                </button>
```
jobara, line 138: Well this is also problematic now because the v2.0 markup and in development one are different. Now we have two panel bars, one for mobile and one for desktop views. We might want to decide what we are documenting in this file exactly. Is it latest, or last stable? If the latter, we may need to change the NPM install instructions or heading.

jhung: Would it be sufficient if we document the stable version, and adding a note about the differences in the latest and linking to another relevant resource?

jobara: I'm actually thinking about this again and I believe we've been documenting what's in master in the general docs. In the future we want to indicate version numbers for changes. What we could/should also include is migration information which we have examples of in the docs.

jhung: Ok. I'll update the doc to reflect what is in master, and add a separate bit of documentation to cover 2.0.

PR #170, opened 2020-03-12
modified the search bar

Pratik10100:


src/layouts/default.html.handlebars
```diff
@@ -92,7 +92,7 @@
                     {{#ifEqual document.category "Search"}}
                     {{else}}
                         <span class="infusion-docs-category">Search</span>
-                        <p class="docs-search-mini"><input type="text" name="qs"/></p>
+                        <p class="docs-search-mini"><input type="text" name="qs" placeholder="search"/></p>
```
apshada: add aria-label="Search Box" and sign cla

jobara:
@Pratik10100 does this address an issue that has been filed? If so, which one?

PR #176, opened 2020-04-03
FLUID-6484: Using an alternative request library

Batey96:
Request has been deprecated by its core and makes sense not to use again in the infusion documentation platform.A similar alternative with same structure and approach Urlib is used instead.
fixes https://issues.fluidproject.org/browse/FLUID-6484


greatislander:
@jobara Makes sense. @Batey96 Thanks for your work on this!

PR #180, opened 2020-04-13
Add flexibility support in "rewriteMdLinks" helper func

Batey96:
This PR fixes https://issues.fluidproject.org/browse/FLUID-6490
Eliminates the use of hand coded *.md files in content.
Now all *.md files can be referenced in any content by just writing the name of the .md file
## Example Usage
```
[view component](tutorial-gettingStartedWithInfusion/ViewComponents.md)
```

The link above can still be referenced just by writing the file name
```
[view component](ViewComponents.md)
```

### DIFF
--- a/docpad.js
+++ b/docpad.js
@@ -26,8 +26,28 @@ var githubDocRoot = "https://github.com/fluid-project/infusion-docs/blob/master/
 // With this helper, we can write links to *.md files in our source files but
 // generate links to *.html in the DocPad output. This arrangement gives us
 // links that work both on the GitHub website and in the generated HTML.
+// For flexibilty,simply write the name of the .md file and the helper will automatically generate the .html link. 
 var rewriteMdLinks = function (content) {
-    return content.replace(/(<a\s[^>]*href="[\w-/\.]+)\.md(["#])/gm, "$1.html$2");
+  var getfullPath =function(filename){
+      var fullPath;
+      var append_parent = "/infusion/"+docsVersion;
+      var stripped_file = path.basename(filename);
+      siteStructure.forEach(site=>{
+        if(site.category!='Search'){
+          site.sections.forEach(section=>{
+              section.pages.forEach(page=>{
+                  if(stripped_file==path.basename(page.href,'.html')){
+                      fullPath = path.join(append_parent,page.href);                          
+                  }         
+              })        
+          }) 
+        }
+     })
+     return fullPath
+}
+     return content.replace(/([\w-/\.]+)\.md(["#])/gm, (match,p1,p2)=>{
+       return getfullPath(p1)+p2
+   }); 
 };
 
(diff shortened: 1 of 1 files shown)


Batey96:
Thanks @YashJipkate for your comments, i did'nt see that coming!. From what your comment, i drew out 3 different concerns
1. Multiple files with same name in different folders:
 What you mean is something like TextfieldStepperAPI.md which is currently categorized as component and found in the document's root to be also found in a place like tutorial-userInterfaceOptions(or some_folder)/TextfieldStepperAPI.md ?
Files are arranged to fall under a particular category and if need be to have component.md in another folder but in thesame category,then i think tagging thesame file with the other category will be the solution.But if we need component.md as a tutorial file or some other category file,then from the current naming conversion in infusion documentation, we could simply append component-tutorials.md or component-some-other-cat.md
2. Build performance
I compared performance for 2 different branches(latest and one_with-the_flex-func)
### with commit
Screenshot from 2020-04-14 19-48-53

### latest
Screenshot from 2020-04-14 19-50-27
 Yes you are right there is an increase in build time
   3. Broken .md links in local;
    The goal for this PR was to target the web: hopefully my content organisation PR will be merged 
    so that developer locally can easily locate the linked file without necessary clicking on the link
 
I might be totally wrong in my statements above:

YashJipkate:
@Batey96 Here are some things I understood from you:

1. There might be a case in the future say we want to include some file like _ChangesFromV1V2.md_ in multiple tut folders - this can be a requirement. Since both of them fall under the 'tutorials' category so the tags would also be the same. Also, the convention has named the **folders** in the way you said, but I couldn't find any files with such a naming convention. Could you give an example of such a file in the current project? Anyways, this heavily relies upon the rigid assumption of file naming and structuring. Don't you think that this kills the idea of 'flexibility' you aim to introduce here? Don't you think the trade-off is too high?

2. The build times don't seem to change significantly, that's a good start, thanks for checking! But the algorithm still looks inefficient in the way it iterates over the whole project structure for every link present in the codebase. I am unsure if this is the best practice.

3. Actually, the links should work in both the GitHub previews as well as the web in my opinion. This makes it easier for developers and content writers to test out the page while editing even in GitHub editor to test out the links via the GitHub preview without waiting for it to be merged or going through the process of local installation and development. Also, the links look more 'logical' when laid out the full path. Imagine a file referenced just by its name and that turns out to reside in some other folder, it would be very difficult to find it. Even if there is a very semantic organization of the files, inter-folder links will exist.

Please feel free to correct me if I'm wrong or misunderstood you.
Thanks!

Batey96:
Hello @YashJipkate very sorry for late response,I have been having unexpected commitments so far.Reason why I have been slow in this repository.(raising issues and popping PRs).Will dedicate time for this your feedback on this PR.thanks

PR #182, opened 2020-04-13
FLUID-6376: Made the site navigation responsive

YashJipkate:
Currently, the top navigation bar is not at all responsive and becomes extremely unable to use on mobile devices and other smaller devices. This PR aims to fix this.

The necessary breakpoints and their respective screenshots are posted in the corresponding [issue](https://issues.fluidproject.org/browse/FLUID-6376)

### DIFF
--- a/src/static/css/infusion-docs.css
+++ b/src/static/css/infusion-docs.css
@@ -632,3 +632,27 @@ Instead have it appear around the enclosed
 .search-result-sub-entry {
     border-top: 1px solid #ffffff;
 }
+
+@media only screen and (max-width: 639px) {
+    .infusion-docs-container .infusion-docs-header {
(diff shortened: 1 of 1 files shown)


coderbabu10:
i think for this we can have an toggle button which would be much more good idea for the mobile users.what you think @jobara

YashJipkate:
@coderbabu10 Using a hamburger would a question of design and need to be discussed with the community. Also, I saw some other sites like [FLOE](https://floeproject.org/) and [fluid](https://fluidproject.org/) and they did not use any hamburger icon but they all made the nav responsive.
But I am open to any design suggestions by the community.

coderbabu10:
Yes @YashJipkate that's why i have tagged @jobara for this.

PR #183, opened 2020-04-29
typo fix:in iocAPI externalisable->externalizable

Chaitanyassr:
fix - issue #184 
typo fix: externalisable->externalizable
Under API in iocAPI.


coderbabu10:
@chaitanyassr i think it is fine with 's' please have a look @justion_o

jobara:
We have contributors from various parts through out the world, so we get documentation and etc with different localized spellings. I don't think we've declared a standardization on which locale to use for spelling. So at this point, I don't think it's something that needs to be changed, but maybe something to think about for the future.

Chaitanyassr:
> We have contributors from various parts through out the world, so we get documentation and etc with different localized spellings. I don't think we've declared a standardization on which locale to use for spelling. So at this point, I don't think it's something that needs to be changed, but maybe something to think about for the future.

Okay no problem, I would be closing down this PR for now.

PR #185, opened 2020-05-20
FLUID-6501: fixed search-docs-mini

sachin10101998:
Fixes FLUID-6501.

### DIFF
--- a/src/documents/to-do/search.md
+++ b/src/documents/to-do/search.md
@@ -1,0 +1,5 @@
+---
+title: Search the Infusion Documentation
+layout: default
+category: Search
+---
--- a/src/documents/tutorial-developerIntroduction/search.md
+++ b/src/documents/tutorial-developerIntroduction/search.md
@@ -1,0 +1,5 @@
+---
+title: Search the Infusion Documentation
+layout: default
+category: Search
+---
--- a/src/documents/tutorial-gettingStartedWithInfusion/search.md
+++ b/src/documents/tutorial-gettingStartedWithInfusion/search.md
@@ -1,0 +1,5 @@
+---
+title: Search the Infusion Documentation
+layout: default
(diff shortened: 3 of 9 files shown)


jobara:
@the-t-in-rtf do you have some time to take a look at this?

the-t-in-rtf:
Sure thing.  I just took a look at [FLUID-6501](https://issues.fluidproject.org/browse/FLUID-6501) for the problem behaviour.  The changes themselves look fairly straightforward, I will verify the fix in local testing and comment back shortly.

the-t-in-rtf:
OK, I can see that the fix is to basically clone the search page into each subdir.  If possible, I would much rather somehow fix the URL used by the mini search in each subdirectory.  So, when running `npm docpad` against your branch and following the steps to reproduce, I end up at this URL:

http://127.0.0.1:9778/infusion/development/tutorial-developerIntroduction/search.html?qs="remarks"

I would expect instead that all the mini search forms point to the same top-level URL, i.e.:

http://127.0.0.1:9778/infusion/development/search.html?qs="remarks"

This can be addressed by using the `relativeUrl` macro to pass in the path to the search form as an option [when instantiating the search form "mini" component](https://github.com/fluid-project/infusion-docs/blob/113cbd7b9ef106c821698fea4f6503d601ba2a02/src/layouts/default.html.handlebars#L163).  You'd just pass the [`targetUrl` option](https://github.com/fluid-project/infusion-docs/blob/113cbd7b9ef106c821698fea4f6503d601ba2a02/src/static/js/mini-search.js#L23).

I'd prefer that to adding a bunch of files, even if they are small.  It also seems more honest, as we don't have targeted searching within the subdirs, which the URL kind of implies to me.

amb26:
Closed for lack of activity

PR #136, opened 2018-04-02
FLUID-6148: Initial pass at docs for new potentia II features

amb26:

PR #187, opened 2020-06-26
Migrate to 11ty.

YashJipkate:
This Pull Request migrates the SSG from Docpad to 11ty.

To test locally: 

- run `npm run 11ty` and open up the local server as pointed by the console output.

Some notes:

- To see a specific version, just put the version name after the baseURL like so: http://<localhost>/infusion/development/**v1.5.0**/ComponentOptionsAndDefaults.html
- The tests don't work anymore. Since they were from docpad era, they need to be migrated to 11ty.


amb26:
I think this work is excellent and am looking forward greatly to the migration to a lively new platform.

Probably some discussion has happened about this already, but I wanted to highlight the discussion around versioning which is sort of a profound question when considering the relationships between a documentary repository, its target code, and its audience. The idea of "versions" in Infusion has sort of been a semi-fiction for a while, although it might cease to be a fiction someday.

What we do seem to need is some kind of relatively fine-grained scheme of associating some development with a particular revision it arose in. Some useful prior art here is the node.js docs, which have two important facilities. The first can be seen everywhere - e.g. at https://nodejs.org/docs/latest/api/process.html#process_event_beforeexit where one can see that this API was "Added in v0.11.2" and so at any point in time one has some historical context. However, in addition to this it is also possible access URLs like https://nodejs.org/docs/latest-v6.x/api/process.html to access the complete docs appropriate for a particular historical revision.

I think both facilities are important, and I'm not proposing that we do anything about this right away, but I'd like people working on this migration to keep this issue knocking around in the back of their minds when structuring the new layout and scripts. The wholesale replication, e.g. of the whole v1.5.0 tree as concrete files, rather than a version snapshot of the existing files, might be a good approach to very distantly historical documentation such as v1.5.0 but might set a confusing precedent for upcoming revisions which are being described in pulls such as https://github.com/fluid-project/infusion-docs/pull/136 which are arguably a description what we might be expecting to be the v3.0.0 release of Infusion.

Any thoughts welcome!

YashJipkate:
Thanks @amb26 for your in-depth insights. In my opinion too, both the features that you highlighted are useful.

I too had some doubts related to what could be the most optimal way to carry out the versioning and started a discussion thread regarding it in fluid-work ([thread](https://lists.idrc.ocad.ca/pipermail/fluid-work/2020-May/011198.html)). While waiting for the views of the community, we decided to go for the subdirectory style versioning (the replication thing) unless proposed with some better approach. It is simple and easy to implement and maintain. It also provides us with the ability to modify the docs of the previous versions. 
The 'added in Vx.y.z' is more of a UI thing, it has to be hardcoded in the doc itself ([like here](https://github.com/11ty/11ty-website/blob/master/docs/events.md#afterbuild--addedin-100-)) and then use a shortcode to style it ([like this](https://github.com/11ty/11ty-website/blob/master/config/addedin.js)). So this feature is something that would be needed to be modified right into the doc files, and the design of the 'Added in' label to be discussed with the community.

I would suggest that we track the 'Added in' label feature with a corresponding issue and after due discussion, it can be resolved easily. 

Also, I don't completely understand what do you mean by the following line:

> The wholesale replication, e.g. of the whole v1.5.0 tree as concrete files, rather than a version snapshot of the existing files, might be a good approach to very distantly historical documentation such as v1.5.0 but might set a confusing precedent for upcoming revisions

Could you please elaborate a bit on it.

Please let me know if there are any suggestions!
Thanks

package.json
```diff
@@ -5,41 +5,32 @@
...
-        "gpii-binder": "1.0.5",
-        "gpii-location-bar-relay": "1.0.4",
-        "infusion": "3.0.0-dev.20190328T144119Z.ec44dbfab",
-        "lunr": "2.3.6",
-        "markdown-it-anchor": "4.0.0",
-        "mkdirp": "0.5.1",
-        "normalize.css": "3.0.1",
-        "octicons": "3.1.0",
-        "urijs": "1.19.0"
+        "@11ty/eleventy": "^0.11.0",
```
the-t-in-rtf, line 15: The versions were previously (and intentionally) all pinned, i.e. they did not use the leading caret to indicate that later versions were acceptable.  I would suggest removing the carets.

the-t-in-rtf:
The linting checks no longer pass, please run `grunt lint` and fix the handful of issues introduced.

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Preferences Framework: Minimal Footprint
+title: "Preferences Framework: Minimal Footprint"
```
the-t-in-rtf, line 2: Why are there quotes here and not in other documents?

the-t-in-rtf:
I tried the versioned URLs, the search is broken in both 1.5.0 and 2.0.0 (the search page does not exist in the versioned directory).  It works for the unversioned copy, but really, if you choose to have versioned directories there is more work to be done than simply copying the search form.  IMO you will need to ensure that:

1. There is some indication of which version you are viewing on every page.
2. There is a clear means of changing versions, which is often in close proximity to the version details.
3. The search results are targeted to the version you're viewing.

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
YashJipkate: Actually this was interfering in the YAML format of the frontmatter (more specifically, the ":"). Hence the qoutes.

package.json
YashJipkate: Done.

YashJipkate:
> The linting checks no longer pass, please run `grunt lint` and fix the handful of issues introduced.

Done.

YashJipkate:
About the versioned search, that is actually somewhat tricky.


> 1. A top-level page with a list of versions that links to each versioned directory.

This would be easier to implement. 

> 2\. A header that indicates the current version being viewed with controls to change versions, including a hot link for "latest".

This would be tricky since according to the current design, I am not storing any version "state". But if I am to use versions in search index then I will have to extract the version from the filePath. Then I already would have a method to get the version so using this method won't harm. Although there would be some design discussions that would be needed to be done with the community.



> Of course, if you follow the approach jQuery and other sites use, and simply add an indication of a version range to each page range, then you won't need to refactor the search, display the selected version, or develop a version list and/or picker.

This is actually a good way to handle versions, I wasn't knowing of it until now. Although this is less intuitive and also I don't quite know what are the exact transition files, I would need to discuss them with the community.

In my opinion, the earlier two options are good and I personally prefer the first one. What do you think?

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
the-t-in-rtf: Thanks, good to know.

jobara:
> What we do seem to need is some kind of relatively fine-grained scheme of associating some development with a particular revision it arose in. Some useful prior art here is the node.js docs, which have two important facilities. The first can be seen everywhere - e.g. at https://nodejs.org/docs/latest/api/process.html#process_event_beforeexit where one can see that this API was "Added in v0.11.2" and so at any point in time one has some historical context. However, in addition to this it is also possible access URLs like https://nodejs.org/docs/latest-v6.x/api/process.html to access the complete docs appropriate for a particular historical revision.
> 

It looks like each page will also only show the versions that actually include the specific document in the version drop down. E.g. https://nodejs.org/docs/latest/api/async_hooks.html That's a really handy feature.

> I think both facilities are important, and I'm not proposing that we do anything about this right away, but I'd like people working on this migration to keep this issue knocking around in the back of their minds when structuring the new layout and scripts. The wholesale replication, e.g. of the whole v1.5.0 tree as concrete files, rather than a version snapshot of the existing files, might be a good approach to very distantly historical documentation such as v1.5.0 but might set a confusing precedent for upcoming revisions which are being described in pulls such as #136 which are arguably a description what we might be expecting to be the v3.0.0 release of Infusion.
> 

Part of the issue we have with docs is that we tend to not finish them by the time of release and may also have typos or other errors that need correcting after the fact. This makes tagging them along with the Infusion release problematic. We could include it as part of the release process. I believe that the docs for node are actually part of the code base. Moving to something like that might help as the docs are actually part to the code and would be more easily tracked along with PRs for code changes. Another suggestion was to keep the docs separate and use branches instead of tags for organizing them, that way we can commit updates and fixes to them as needed. 

Another issue we have with the docs is that they are highly coupled to the SSG that they are run with. So we've tagged versions of the docs in the past, but those won't work with 11ty. And even more distant versions of the docs are probably just in the wiki, but maybe those are sufficiently old that we don't need to worry about them. The branch suggestion above would allow us to restructure the versioned docs to work with 11ty or any future SSG. In the original GSoC project idea for this work, we were going to explore completely decoupling the docs content from the SSG so that they could be more easily migrated between tools as needed. That's actually something we may still want to look at, for example if we move the docs into the Infusion repo itself or want to be able to produce different outputs e.g. ebook, pdf, etc.

YashJipkate:
@jobara what do you have in mind regarding displaying the older versions? I mean should we add just the "added in"/"removed in" labels (basically a version range) and do away with version-based displaying of the pages?

jobara:
@YashJipkate I think we can start with that approach, that is using labels within the page. In the future we can look at displaying specific versions separately and layer that in on top of the labelling approach.

YashJipkate:
@jobara so how should we carry out the process of labelling. Is there a way or a source from where I can see which features need which labels? 

What I think that we should merge this PR with the provision of the labelling feature and label some (4-5) pages as a sample. And then we can open an issue to track the labelling of the remaining pages. This issue can be solved incrementally since doing it in one go is practically impossible given the number of pages.

What do you think?

jobara:
@YashJipkate Those are good points. It would be good to have a sample page to set the precedent for how we want to label things and some documentation about that as well. Perhaps @amb26 has some thoughts on what might serve as a good example page and could provide some assistance on the correct labelling. Also @amb26 it would be good to hear your thoughts on this in general.

YashJipkate:
@jobara while @amb26 answers them, could you tell what are your insights on how to label. We can make a draft page design and modifications can be done to it as we get more suggestions.
Or should we first merge this PR sans the versioning and add it in another PR? This way we can configure and test the newly migrated site in the production environment since there might be some problems in it as I don't have much knowledge of the hosting and hence haven't tested it.
Once that is stabilized, addition of versioning would be just another feature that can be easily integrated.

What do you think?

jobara:
@YashJipkate lets roll back the versioning. So no subdirectory and etc. Just a straight port to start. Could you also check in with @sachin10101998 to compare the migration plan between this site and the others.

src/documents/tutorial-prefsFrameworkMinimalFootprint/MinimalFootprint.md
sachin10101998: Use `&#58;` in place of colon in title.

YashJipkate: This doesn't seem to work, the title shows literal `&#58;`.

YashJipkate:
@jobara I have rolled back the versioning. Could you please take a final look before we test it out in production environment?
@sachin10101998 Please see if this conforms to the migration plan of yours.

sachin10101998:
@YashJipkate  I'll review it by the weekend and suggest changes (if any) to keep different sites of fluid in sync. I am also in the process of creating a starter kit for 11ty which follows the same protocol as floe and fluid project. Will update you in a couple of days.

YashJipkate:
@sachin10101998 Do you have any reviews/suggestions?
Also, could you please look if this can be deployed to production successfully. I am somewhat skeptical about the pathPrefix [here](https://github.com/fluid-project/infusion-docs/pull/187/files#diff-ec92918f71da79b99bf95abac6a52650R56).

package.json
```diff
@@ -5,30 +5,21 @@
     "license": "CC-BY-3.0",
     "repository": "https://github.com/fluid-project/infusion-docs",
     "scripts": {
-        "docpad": "node node_modules/docpad/bin/docpad run --env static --out out",
-        "deploy": "node node_modules/docpad/bin/docpad deploy-ghpages --env static",
-        "pretest": "node node_modules/docpad/bin/docpad generate --env static --out out",
+        "11ty": "mkdir -p out && eleventy --serve",
```
sachin10101998, line 8: You don't need to specify the mkdir. You have already specified the out folder in ".eleventy.js" file and putting it again here may cause consistency issues in case output directory is changed in future. chane it to "11ty": "eleventy --serve", . similar for build script.

package.json
```diff
@@ -5,30 +5,21 @@
...
-        "docpad-plugin-handlebars": "2.4.0",
-        "docpad-plugin-highlightjs": "2.5.0",
-        "docpad-plugin-markit": "2.4.0",
+        "@11ty/eleventy": "0.11.0",
         "foundation-sites": "5.5.2",
-        "fs-extra": "4.0.2",
         "gpii-binder": "1.0.5",
         "gpii-location-bar-relay": "1.0.4",
-        "infusion": "3.0.0-dev.20190328T144119Z.ec44dbfab",
+        "highlight.js": "10.0.3",
```
sachin10101998, line 19: Shift the dependencies which are not required for production to devDependencies.

package.json
```diff
@@ -5,30 +5,21 @@
...
+        "infusion": "3.0.0-dev.20200326T173810Z.24ddb2718",
```
sachin10101998, line 20: infusion is just being copied through makethroughCopy of 11ty. So we can put it under dev dependencies too.

package.json
YashJipkate: I did this for the search index file. It is being generated in the output directory and previously the output directory was produced before the generation of the search index but now the output directory is produced after the generation and the path fails to resolve if there is no out/ directory already.

sachin10101998:
The site looks good and is almost close to its destination but I had a few things to mention:
1) Site structure.json is present in the root directory as well as the src/data folder too. any special reason for having multiple copies of the same?
2) Javascript code sat many places overlap. Here's a sample screengrab of what I am talking about. 
Screenshot from 2020-07-28 01-07-15
3) Transfer the dependencies which are not required for production to devDependencies. You can copy infusion to dev Dependencies too since it's just copied through passthroughcopy of 11ty.
4) No Inline HTML should be present in MD posts. It's better to switch them to HTML rather than MD. It's not going to make any difference and that's the solution we found for the issue in floe and fluid. Am talking about all the MD pages. However, I see that you have converted a few tags to MD. We can discuss this again @BlueSlug @jobara @jhung @colinbdclark . It's not a good practice to have inline HTML. In fact, my markdown extension shows this as warning. Here's a screengrab. 
Screenshot from 2020-07-28 01-18-26
5) Also, @gtirloni can the docker image be built for infusion docs the same way we built them for floe and fluid?

package.json
YashJipkate: Are you sure about this? Aren't devDeps not installed in the production server? So 11ty won't have a folder to make a passthrough copy and infusion is used extensively. Are you sure all of those usages are dev-specific? If yes, then I'll add them to devdeps. Also, it was originally in the main deps.
It may be that I have a lesser understanding of this.

sachin10101998: I tested it on local and it seems to be working.

sachin10101998: We had a similar discussion and @jhung suggested this.

YashJipkate:
> Site structure.json is present in the root directory as well as the src/data folder too. any special reason for having multiple copies of the same?

Only the one at data/ is needed, I think I forgot to drop the root one. Removed.

> Javascript code at many places overlap. Here's a sample screengrab of what I am talking about.

This happens because of the poor rendering of the `<pre><code>` blocks by the MD template engine of 11ty. I overcame this by converting the `<pre><code>` to `<pre class="highlight"><code class="hljs javascript">` so highlights could instead render it. Thanks for pointing out!
Screenshot from 2020-07-28 02-00-54

> However, I see that you have converted a few tags to MD.

I don't understand? I have not converted anything. Are you trying to say something else?

> We can discuss this again @BlueSlug @jobara @jhung @colinbdclark. It's not a good practice to have inline HTML. In fact, my markdown extension shows this as warning.

Didn't know this. Thanks for the info. Suggestions are welcome!

package.json
YashJipkate: Okay I'll shift infusion to devdeps. Thanks for checking.
Also, are there other packages that are required to be shifted to devdeps. Since I mistook infusion, I may be wrong for others too.

amb26:
> We can discuss this again @BlueSlug @jobara @jhung @colinbdclark. It's not a good practice to have inline HTML. In fact, my markdown extension shows this as warning.

It may not be a good practice, but there are several examples in the docs of tables which are too complex to express in markdown, so we'll need to retain this capability:
https://github.com/fluid-project/infusion-docs/blob/master/src/documents/IoCTestingFramework.md#supported-fixture-records

Do ensure that any new files that are added to the project are caught by the linting patterns in Gruntfile.js

YashJipkate:
> Do ensure that any new files that are added to the project are caught by the linting patterns in Gruntfile.js

We can have a Github actions for that, similar to the CLA check. So any new changes would be first tested by it. What do you think?

Also, @sachin10101998 could you address the comments above and take one final look.

Thanks!

package.json
YashJipkate: > I tested it on local and it seems to be working.

Are you sure you ran `npm install --production` because running the same in my local didn't get infusion.

jobara:
> We can have a Github actions for that, similar to the CLA check. So any new changes would be first tested by it. What do you think?

Yes, we should add a CI check for that, but it's also good to run those checks locally before pushing to a branch and etc.

package.json
sachin10101998: Floe and fluid have been deployed using 11ty and both have infusion in devdependencies .sites are working fine

package.json
```diff
@@ -5,30 +5,21 @@
     "license": "CC-BY-3.0",
     "repository": "https://github.com/fluid-project/infusion-docs",
     "scripts": {
-        "docpad": "node node_modules/docpad/bin/docpad run --env static --out out",
-        "deploy": "node node_modules/docpad/bin/docpad deploy-ghpages --env static",
-        "pretest": "node node_modules/docpad/bin/docpad generate --env static --out out",
+        "11ty": "mkdir -p out && eleventy --serve",
+        "build": "mkdir -p out && eleventy",
```
jhung, line 9: `mkdir` does not perform consistently across operating systems (i.e. Windows).

Use npm package `rimraf` instead to create the directory if needed. See this as an example: https://github.com/fluid-project/floeproject.org/blob/53159c4e8db1e5705e3d0e54569e353be0f2da0b/package.json#L8

Also, it would be good to have the npm scripts be consistent with other IDRC web projects (i.e. `start` `build` and `clean` scripts). For example:
* [IDRC website package.json](https://github.com/inclusive-design/idrc/blob/dev/package.json)
* [Floe Project Website package.json](https://github.com/fluid-project/floeproject.org/blob/master/package.json)

YashJipkate: I have now used `mkdirp` to do the work of `mkdir -p`.

I have now renamed the scripts as per the two references given by you. `clean` uses the `rimraf` package, whereas `start`, `serve` and `build` don't need it since `mkdirp` overwrites if the directory is already present.

YashJipkate:
Hi @jobara @gtirloni could you please help me with the build failures. I tried changing the Dockerfile too but that doesn't seem to work. Is it some script in the Netlify configuration?
Please take a look.

Thanks!

jobara:
@greatislander do you have thoughts on what the issue might be?

gtirloni:
@YashJipkate could you add a netlify.toml file to the root directory that adds the proper build commands? See an example here: https://github.com/inclusive-design/wecount.inclusivedesign.ca/blob/dev/netlify.toml

YashJipkate:
Hi @jobara @gtirloni I have fixed the build now. Please take a look and let me know if any more changes are required.

Thanks!

greatislander (changes requested):
@YashJipkate Thanks for your ongoing work on this! I think the main piece that's missing here is the contents of [src/ghpages-files](https://github.com/YashJipkate/infusion-docs/tree/migrate-ssg/src/ghpages-files). These should be copied without modification to the root of the build directory— this will handle the redirect to `/infusion/development/` that you were trying to handle with the Netlify redirect configuration, as well as adding a redirect from `/infusion/` to `/infusion/development/` and providing a default 404 page. Can you try to integrate this into the build process? Thanks!

YashJipkate:
@greatislander I have moved the contents of src/ghpages-files to the root as you requested. Could you please take a look?

greatislander:
@YashJipkate It looks like you are still using a Netlify redirect which isn't ideal; it means that visiting https://deploy-preview-187--fluid-infusion-docs.netlify.app and https://deploy-preview-187--fluid-infusion-docs.netlify.app/infusion/ and https://deploy-preview-187--fluid-infusion-docs.netlify.app/infusion/development/ all show the same page, which will cause a lot of confusion for search indexing. Is it not possible to copy those files to the root of the `/out` directory?

YashJipkate:
@greatislander I have now implemented your suggestion. Seems that I was not properly handling the ghpages. Please take a look now.

Thanks!

src/layouts/default.hbs
```diff
@@ -61,52 +61,52 @@
...
                                     <ul>
                                         {{#each pages}}
                                             <li>
-                                                {{#ifEqual href ../../../../document.url}}
+                                                {{#ifEqual href ../../../page.url}}
                                                     <span class="infusion-docs-TOC-pageName">{{pageName}}</span>
                                                 {{else}}
-                                                    <a href="{{relativeUrl href ../../../../../document.url}}"><span class="infusion-docs-TOC-pageName">{{pageName}}</span></a>
+                                                    <a href="{{ url href }}"><span class="infusion-docs-TOC-pageName">{{pageName}}</span></a>
                                                 {{/ifEqual}}
```
greatislander, line 110: This appears to be broken. For reference, compare:

Old version: https://docs.fluidproject.org/infusion/development/understandinginfusioncomponents
New version: https://deploy-preview-187--fluid-infusion-docs.netlify.app/infusion/development/understandinginfusioncomponents

Note that in the old version, the "Understanding Infusion Components" link in the left sidebar navigation is turned into a span and highlighted to indicate the current page. There's no differentiation from the other sidebar links in the new version.

greatislander (changes requested):

src/layouts/default.hbs
YashJipkate: This appears to have crawled in some commit previously because I specifically remember taking care of this. Anyway, thanks for noticing, I have fixed them now. Please take a look.

greatislander (approved):
This looks good to me as an initial migration, but others should review again as well, I think.

PR #190, opened 2021-05-05
feat: replace Docpad with Eleventy (#187)

### LINKED ISSUE #151: Activate hover event to show icon while tabbing
I ran the current master of infusion docs. While tabbing through the doc pages, the links for headings looked something like this-
Screenshot from 2020-02-23 08-22-21
It feels out of place here.

Hovering over the heading using the cursor looked something like this:
Screenshot from 2020-02-23 08-22-30
While here it is apparent that the link refers to the heading.

Doing something similar to :hover mouse-based event while tabbing would remove the ambiguity from the tab selected link.

### LINKED ISSUE #156: error with show preference
When we go to a new page it shows show display preference for some time but after it disappears and only show preference is shown 
ezgif com-video-to-gif(4)

i would like to take this issue

### LINKED ISSUE #164: Hovering action for link
**Issue**
Contrast does not have any hovering action on Link as shown above

greatislander:
Picking up on @YashJipkate's excellent work in #187.

So far this PR:

- [x] Replaces `fluid-grunt-lint-all` with `fluid-lint-all`
- [x] Simplifies the build process while preserving the existing directory structure
- [x] Improves search field accessiblity ([FLUID-6476](https://issues.fluidproject.org/browse/FLUID-6476))
- [x] Show header anchor link icons when they are focused, adds accessible label to header anchor links (resolves #151)
- [x] Sets UIO panel button text to a consistent label so it doesn't change when UIO is initialized (resolves #156)
- [x] Adds link hover styles for some contrast themes which lacked them (resolves #164)


.eleventy.js
```diff
@@ -0,0 +1,76 @@
...
+        "node_modules/gpii-location-bar-relay/src/js": "lib/gpii-location-bar-relay",
+        "src/static/css": "css",
+        "src/static/fonts": "fonts",
+        "src/static/js": "js",
+        "src/scripts": "js",
+        "src/documents/images": "images",
+        "src/icons": "/"
+    });
+
+    eleventyConfig.addHandlebarsHelper("ifEqual", function (a, b, options) {
```
amb26, line 52: Could we also migrate over the original comments that accompanied these helpers in the old docpad.js?

amb26:
This is great work - and the build actually works for me out of the box on Windows which it never did before, as well as running much more quickly.
There is one minor linting failure - 

12:30:10.823:    - src\documents\TextToSpeechAPI.md:
12:30:10.825:      (126:-) MD013:line-length - Expected: 120; Actual: 128

which suggests that we need to add the linting task to the CI check

amb26:
Before we can push out a release based on these docs we need to complete a migration from the outdated gpii-era dependencies to the newer versions maintained by @the-t-in-rtf - 
e.g. gpii-testem -> fluid-testem, gpii-binder -> fluid-binder, gpii-express -> fluid-express

but this can be done as part of a separate pull that targets all outdated dependencies if we can round off this one. However, we should the existing tests as executed via "npm run test" to the CI configuration to ensure that they haven't regressed.

amb26:
This pull also needs a README update since these instructions still refer to "npm run docpad"

greatislander:
Hi @amb26 — I ran the node and browser tests locally and they all passed. The CI does include linting but it's not running yet because the GitHub Actions configuration is in this branch, not the target (main). I just fixed that Markdown line length issue so that should be good now.

I'll open another issue to capture the gpii- to fluid- package updates. And you're absolutely right, I need to update the README. Will do that right away.

amb26:
Two small regressions with respect to what is currently in main:

i) The hover anchors used to appear (e.g. in https://docs.fluidproject.org/infusion/development/functiongrades ) when the mouse was hovered anywhere over a heading text. They now only appear if the mouse is hovered in a region somewhere to the left of the text, which is an interaction that the user may not easily discover

ii) The old README used to contain instructions for the task to deploy the updated documentation to the live site at https://docs.fluidproject.org via "npm run deploy" - it looks like these instructions were already outdated, but we should include a section in the README that clarifies that such deploys now automatically execute via Netlify for each merge to main - or just in general explains to the reader about Netlify idioms, e.g. that there is a deploy preview available on every pull request etc. with a link to https://docs.netlify.com/

greatislander:
@amb26 Thanks for the comments, I've addressed both of these!

PR #196, opened 2021-05-07
feat: integrate Eleventy Navigation plugin

greatislander:
I've configured the main navigation to use Eleventy's navigation system. Replacing the docs manifest file poses some issues because some items appear as the children of multiple sections (e.g. the DOM Binder API which appears under both 'API' and 'Views'). I think it would be worth examining the structure prior to any further changes (and I still think embedding navigation/TOC structure in documentation makes more sense than maintaining a parallel manifest file) — however, this PR is now ready for review and merge as is.


src/_includes/layouts/base.njk
```diff
@@ -0,0 +1,165 @@
...
+                    {% for item in navItems %}
+                        <a href="{{ item.url }}"{% if page.url === item.url %} aria-current="page"{% endif %}{% if item.key === category%} class="infusion-docs-category-active"{% endif %}>{{ item.key }}</a>
+                    {% endfor %}
+                </nav>
+            </header>
+
+            <div class="infusion-docs-mainBody">
+                <nav id="tableOfContents" class="infusion-docs-TOC small-12 medium-3 column" tabindex="-1">
+                    {% if category === "Search" %}
+                    {% else %}
```
cindyli, line 89: Rather than having an empty `if` block, shall these two lines be replaced with `{% if category !== "Search" %}`?

Issue #195, opened 2021-05-06
Use Eleventy Syntax Highlight plugin instead of HighlightJS


greatislander:
## Is your feature request related to a problem?

Currently, we use a custom integration of HighlightJS to highlight code blocks: https://github.com/fluid-project/infusion-docs/blob/main/.eleventy.js#L21-L31

Eleventy has a great syntax highlighting plugin that would make this much cleaner and is used in some other projects e.g. the ILDH site: https://github.com/fluid-project/docs-inclusive-learning/blob/main/.eleventy.js#L18

## Describe the solution you'd like

Replace the custom HighlightJS integration with the Eleventy Syntax Highlighting plugin, adapt the stylesheets used for HighlightJS markup to support Prism as seen in ILDH (https://github.com/fluid-project/docs-inclusive-learning/blob/main/src/assets/styles/partials/prism.styl)

## Describe alternative solutions you've considered

Not applicable.

## Additional context or notes

Not applicable.

jobara:
When this is updated, we should make sure that all of the syntax highlighting support the required contrast ratios.

PR #197, opened 2021-05-07
fix: resolve colour contrast issues (fix #192)

### LINKED ISSUE #192: Review contrast throughout site
## Describe the bug

As identified by @jobara (https://github.com/fluid-project/infusion-docs/issues/152#issuecomment-832181179), some content lacks an accessible contrast ratio. We should undertake a comprehensive review of contrast throughout the site, perhaps starting with an Axe audit.

greatislander:
Fixes the contrast issues related to link colour (normal and hover states).

### DIFF
--- a/src/static/css/infusion-docs.css
+++ b/src/static/css/infusion-docs.css
@@ -69,7 +69,7 @@ td {
 .fl-prefsEditor-buttons button:focus,
 .infusion-docs-TOC li a:focus .infusion-docs-TOC-pageName,
 a:focus {
-    outline: solid 0.2rem #008cba;
+    outline: solid 0.2rem hsl(195, 100%, 31%);
 }
 
 .infusion-docs-header a:focus,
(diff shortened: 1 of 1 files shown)


src/static/css/infusion-docs.css
```diff
@@ -69,7 +69,7 @@ td {
 .fl-prefsEditor-buttons button:focus,
 .infusion-docs-TOC li a:focus .infusion-docs-TOC-pageName,
 a:focus {
-    outline: solid 0.2rem #008cba;
+    outline: solid 0.2rem hsl(195, 100%, 31%);
```
jobara, line 72: Out or curiosity, why the switch to hsl? Also I'd assume you'd typically do these types of changes with CSS custom properties. Are you thinking of leaving that for a larger reworking of the entire stylesheet?

greatislander: Switch to HSL was so that I could darken the current hue without changing its other properties (it was `hsl(195, 100%, 35%)` prior to this change). And yes, I don't think making any structural changes to the stylesheet makes sense at this point— there's a lot to consider.

jobara:
@greatislander I think some of the syntax highlighting in the code blocks may also not meet required contrast levels. I've left a note on the issue and comment in #195 to address when we switch the highlighting mechanism.

jobara:
Merged at 38608da32eeb524281067e285efcd899bc641e77

PR #207, opened 2022-02-15
FLUID-6719: Added migration notes about minified bundles.

the-t-in-rtf:
See [FLUID-6719](https://issues.fluidproject.org/browse/FLUID-6719) for details.

### DIFF
--- a/src/documents/DeprecatedIn3_0.md
+++ b/src/documents/DeprecatedIn3_0.md
@@ -40,6 +40,21 @@ The following files and all their contents are deprecated:
 * src/framework/renderer/js/fluidParser.js
 * src/framework/renderer/js/fluidRenderer.js
 * src/framework/renderer/js/RendererUtilities.js
+* dist/infusion-all-no-jquery.min.js
+* dist/infusion-all-no-jquery.min.js.map
+* dist/infusion-all.min.js 
+* dist/infusion-all.min.js.map 
+* dist/infusion-framework-no-jquery.min.js 
+* dist/infusion-framework-no-jquery.min.js.map 
+* dist/infusion-framework.min.js 
+* dist/infusion-framework.min.js.map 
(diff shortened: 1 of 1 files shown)


jobara:
@the-t-in-rtf, I guess these changes are more than deprecations because those files have been removed. It might be better expressed in the [API Changes from 2.0 to 3.0](https://docs.fluidproject.org/infusion/development/apichangesfrom2_0to3_0) guide instead.

the-t-in-rtf:
I agree that the wording could be improved, but they're not API changes either.  I'll try fleshing out the wording a bit so that it's clear that they're "deprecated and deleted".

amb26:
ok with me

src/documents/DeprecatedIn3_0.md
```diff
@@ -41,6 +41,24 @@ The following files and all their contents are deprecated:
...
+* dist/infusion-framework-no-jquery.min.js.map
+* dist/infusion-framework.min.js
+* dist/infusion-framework.min.js.map
+* dist/infusion-uio-no-jquery.min.js
+* dist/infusion-uio-no-jquery.min.js.map
+* dist/infusion-uio.min.js
+* dist/infusion-uio.min.js.map
+
+We only produce minified javascript bundles now, so if you previously used one of the above minified files, you
+can simply update the relative path/filename to remove the `min`, i.e. `dist/infusion-all.js`.
```
jobara, line 60: The section that these are in, is actually a sub section of Renderer. This should probably be in a new Bundles or Builds section that would be a sibling to Pager, Renderer and Fast XML Pull.

the-t-in-rtf: Good catch, I moved it to its own section.  It's now similar to the pager section, in that it has a "bundles" heading and then details about "files".

PR #208, opened 2022-02-15
GH-193: Update dependencies, especially fluid-X ones (resolves #193).

### LINKED ISSUE #193: Update gpii- packages to fluid- equivalents
## Is your feature request related to a problem?

As per https://github.com/fluid-project/infusion-docs/pull/190#issuecomment-833453893, we should replace the gpii packages used for testing this project with their fluid equivalents.

## Describe the solution you'd like

See above.

## Describe alternative solutions you've considered

N/A.

## Additional context or notes

N/A.

the-t-in-rtf:
See #193 for details.


.eleventy.js
```diff
@@ -30,8 +30,12 @@ module.exports = function (eleventyConfig) {
             return "<pre class=\"highlight\"><code class=\"hljs " + lang + "\">" + markdownit.utils.escapeHtml(str) + "</code></pre>";
         }
     });
+
+    var string = require("string");
```
greatislander, line 34: It may make more sense to use the `slugify` package that Eleventy uses for URL generation (see: https://www.11ty.dev/docs/filters/slug/). Note that in Eleventy 1.0, which this site doesn't use yet, they are migrating to `@sindresorhus/slugify`, so maybe adding that dependency and using it will make the Eleventy upgrade cleaner when the time comes).

greatislander (changes requested):

package.json
```diff
@@ -26,7 +26,7 @@
         "lunr": "2.3.9",
         "markdown-it-anchor": "8.4.1",
         "octicons": "3.1.0",
-        "string": "^3.3.3"
+        "string": "3.3.3"
```
greatislander, line 29: Can this be removed?

greatislander (changes requested):

package.json
the-t-in-rtf: Done.

.eleventy.js
the-t-in-rtf: As discussed in the chats, slugify is an ES6 module that doesn't play very nicely with our existing approach.  We decided to use the older slugify built into our version of eleventy.

amb26:
I tested this out and it linted and tested ok - there are some fairly stale Infusions in the overall tree but we can deal with this after Infusion 4.0.0 is released. Could you please add "IE" to the skipped browsers in testem.js since Infusion no longer works with it - cheers!

the-t-in-rtf:
@amb26, I added IE to the list of skipped browsers.

Issue #209, opened 2022-02-15
Browser tests failing.


the-t-in-rtf:
The browser tests fail in `main`, and likely have been failing for a while.  I will make a go at fixing them, as:

1. It's not great to have no way to verify changes are working other than manual testing.
2. I really want us to start running them in CI.

PR #212, opened 2022-02-16
GH-210: Added tests to CI builds (resolves #210).

### LINKED ISSUE #210: Run tests in CI.
Currently we run linting checks and confirm that the site builds without errors.  Once #209 is resolved, we should also run our tests in CI.

the-t-in-rtf:
See #210 for details.


the-t-in-rtf:
This should work independently of the fix in #209, as it calls the tests after the build has already been run.

the-t-in-rtf:
Not sure why the workflow actions didn't run.

the-t-in-rtf:
Thanks to @greatislander we figured out that there was an extra space on one line in the config.  Fixing that got everything working.  This should be ready for review.

.eleventy.js
```diff
@@ -30,8 +30,12 @@ module.exports = function (eleventyConfig) {
             return "<pre class=\"highlight\"><code class=\"hljs " + lang + "\">" + markdownit.utils.escapeHtml(str) + "</code></pre>";
         }
     });
+
+    var string = require("string");
```
greatislander, line 34: It may make more sense to use the `slugify` package that Eleventy uses for URL generation (see: https://www.11ty.dev/docs/filters/slug/). Note that in Eleventy 1.0, which this site doesn't use yet, they are migrating to `@sindresorhus/slugify`, so maybe adding that dependency and using it will make the Eleventy upgrade cleaner when the time comes).

greatislander (changes requested):

.eleventy.js
the-t-in-rtf: As discussed in the chats, the newer slugify requires ES6, which is way too much work for this effort.  We agreed to use the version bundled with our (old) version of eleventy, and I've confirmed that works fine.

Issue #225, opened 2022-03-14
A single header id in the docs does not match what GitHub generates


jobara:
## Describe the bug

The heading for fluid.log in CoreAPI.html produces a different anchor on GitHub than in our rendered docs.

The text of the heading is: `fluid.log([logLevel, arg1, ... argn])`.  GitHub generates the header ID as "fluidlogloglevel-arg1--argn".
The new combination of markdown-it and [markdownit-anchor](https://www.npmjs.com/package/markdown-it-anchor) generates the ID as "fluidlogloglevel-arg1-argn" (note the single dash between "arg1" and "argn").

### To reproduce

Link on the docs site: https://docs.fluidproject.org/infusion/development/coreapi#fluidlogloglevel-arg1-argn
Link in GitHub: https://github.com/fluid-project/infusion-docs/blob/main/src/documents/CoreAPI.md#fluidlogloglevel-arg1--argn

### Expected behavior

The header IDs are the same so that the page anchors will be identical. This is important so that internal references will work when viewing the documentation on the website or in GitHub.

## Additional context or notes

Originally filed as [FLUID-6207](https://issues.fluidproject.org/browse/FLUID-6207)

amb26:
Do we have a route to doing something about this? It seems like it might require forking a number of packages. I'd generally resigned myself to having internal references not work within GitHub which as the original issue reports has not worked for years.

jobara:
@amb26 as you mention one option is to not enforce/expect internal references to work on GitHub. And then we just mark this as won't fix and close it. 

Another thing to consider is if we want to keep maintaining links with Markdown files. At the moment we do some processing to convert the links to point at the related html file during build. I suppose a similar step could be taken for this case where we do some processing at build to replace `--` with `-`.

greatislander:
I saw this issue and found an easy way to resolve it without any post-processing or other hacks via the [github-slugger](https://github.com/Flet/github-slugger) package.

PR #230, opened 2022-03-15
GH-225: align slugs with GitHub’s algorithm

greatislander:
Replaces bundled slugify package with github-slugger, which generates slugs for anchors using the same algorithm that GitHub uses. Resolves GH-225.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -11,13 +11,10 @@
 */
 "use strict";
 
-var slugify = require("slugify");
-var cleanSlugify = function (string) {
-    return slugify(string, {
-        replacement: "-",
-        lower: true,
-        strict: true
-    });
+var GithubSlugger = require("github-slugger");
+var githubSlugify = function (string) {
+    var slugger = new GithubSlugger();
+    return slugger.slug(string);
 };
 
 require("./index.js");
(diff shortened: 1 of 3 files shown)


the-t-in-rtf:
The minimal code changes look reasonable.  I did some QA to verify that:

1. The updated links work within the preview site
2. The updated slugs match those used to link to [a section within the markdown on GitHub](https://github.com/fluid-project/infusion-docs/blob/4bc1128b1984102655c07996db7d765e569738e6/src/documents/CoreAPI.md#fluidfailarg1--argn).

As it had been a while since I worked with it, I confirmed that the link checker programatically checks header links as well, so we can be confident that all the slugs we have on the site are usable.

.eleventy.js
```diff
@@ -11,13 +11,10 @@
...
-    return slugify(string, {
-        replacement: "-",
-        lower: true,
-        strict: true
-    });
+var GithubSlugger = require("github-slugger");
+var githubSlugify = function (string) {
+    var slugger = new GithubSlugger();
+    return slugger.slug(string);
 };
```
jobara, line 18: I think you can simplify this down by just using the internal implementation instead of constructing a new instance. While using the constructor is recommended, it appears that the current implementation would create a new instance for each call, likely negating benefits of the single instance. In some local testing, this still produces the desired slugs and increments for duplicate headings. You can look at the "Example:" headers used on DataSourceAPI.md and ConnectingThePartsOfAPreferencesEditor.md. 

You could hoist the `var slugger = new GithubSlugger();` out of the function definition to have a single instance. However, if I'm not mistaken, this would cause duplicate slugs across pages to increment the suffix counter. Which I don't believe we'd want. 

see: https://github.com/Flet/github-slugger/blob/master/README.md

```suggestion
var githubSlugify = require("github-slugger").slug;
```

greatislander: Makes sense, thanks!

jobara (approved):

PR #241, opened 2022-06-24
GH-240: Update and reorganise docs on event injection to make preferred approached clearer

### LINKED ISSUE #240: Add warning notice to use of wholesale event injection
As encountered by @duhrer in Matrix today https://matrix.to/#/!JcoHDrfLedPQdFhAXn:matrix.org/$wWtU3ndoBRsNT8mlAIrvj746pEaaIVF6mGpqe-nGVuE?via=matrix.org the wholesale event injection pattern advertised on https://docs.fluidproject.org/infusion/development/eventinjectionandboiling#event-injection can lead to subtle bugs when listeners with the same namespace are attached to incarnations of "the same event" as it appears around the tree - these will unexpectedly become deduplicated.
This issue has been known for a while and should be warned against in the docs.

amb26:
Resolves #240

### DIFF
--- a/src/documents/EventInjectionAndBoiling.md
+++ b/src/documents/EventInjectionAndBoiling.md
@@ -81,9 +81,9 @@ that.events.parentEvent.fire(that, 5); // strikes above listener through injecte
 
 <div class="infusion-docs-note">
 
-<strong>Note:</strong> Both of these kinds of boiling can be applied at the same time - that is, it is possible to adjust the
-signature of a listener using `args` at the same time as resolving to it elsewhere in the tree by means of an IoC
-reference key. Note also that all injected listeners automatically deregistered by the framework when the component
+<strong>Note:</strong> Both of these kinds of boiling can be applied at the same time - that is, it is possible to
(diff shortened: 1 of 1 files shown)


src/documents/EventInjectionAndBoiling.md
```diff
@@ -9,10 +9,82 @@ component. Within a larger design, sometimes it is necessary to
...
+
+var that = examples.injectedListenerParent();
+that.events.parentEvent.fire(that, 5); // strikes above listener through injected listener attachment
+```
+
+_Note that both of these kinds of boiling can be applied at the same time - that is, it is possible to adjust the
+signature of a listener using `args` at the same time as resolving to it elsewhere in the tree by means of an IoC
+reference key. Note also that all injected listeners automatically deregistered by the framework when the component
+which holds their record (e.g. the `child` component in this example) is destroyed - there is no need for the user to
+call `removeListener` manually._
```
jobara, line 86: These should probably use our Note styling/syntax.

Issue #246, opened 2022-10-19
Update documentation for migrating from infusion 4.0 to 4.5


cindyli:
## Is your feature request related to a problem?

Infusion 4.5.0 will be released soon. Update the infusion documentation site to reflect changes in 4.5.0.

jobara:
Appears to have been resolved at https://github.com/fluid-project/infusion-docs/commit/171ad71e9eae54bd208110ce2201da5079f254dc

PR #247, opened 2022-10-21
GH-246: Add doc for infusion 4.5.0 release and update package dependencies

cindyli:
This pull request does:

1. Add documentation for the infusion 4.5.0 release;
2. Update package dependencies. The only package that is not yet upgraded is "foundation-sites". The infusion doc site uses css files from this package as base styles. As the new release of "foundation-sites" no longer includes "normalize.css"  and has a bunch of styling changes in itself, upgrading it requires further styling fixes on the doc site. If we decide to upgrade this package, it probably worth a separate ticket.


src/documents/APIChangesFrom4_0To4_5_0.md
```diff
@@ -0,0 +1,63 @@
...
+
+### UI Options Template In Infusion 4.0
+
+```html
+<body>
+    <!-- BEGIN markup for Preference Editor -->
+    <section class="flc-prefsEditor-separatedPanel fl-prefsEditor-separatedPanel">
+        <!--
+            This div is for the sliding panel bar that shows and hides the Preference Editor controls in the mobile view.
+            A separate panel bar for mobile displays is needed to preserver the correct tab order.
```
amb26, line 30: typo preserver -> preserve

src/documents/APIChangesFrom4_0To4_5_0.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+    <!-- the rest of your page here -->
+    <h1>My Website</h1>
+</body>
+```
+
+### UI Options Template in Infusion 4.5
+
+Remove the css selector `flc-prefsEditor-iframe` from the div that will contain the Preference Editor component.
```
amb26, line 63: I think it might confuse people to only quote the markup which is now obsolete and give them an instruction how to correct it - perhaps better to quote the corrected markup in full?

src/documents/IoCReferences.md
```diff
@@ -44,7 +44,7 @@ the meaning and form of the context name can vary and have a different meaning i
                             <li>
                                 an entire <code>&lt;reference&gt;</code> string in itself. This is useful in highly
                                 dynamic cases where the context name should be determined from some other options
-                                material, e.g. <code>{{that}.options.targetContext}</code>
+                                material, e.g. <code>{% raw %}{{that}.options.targetContext}{% endraw %}</code>
```
amb26, line 47: What is the %raw% directive issued against? Does the {{ confuse the markdown processor somehow?

cindyli: Yes, exactly. The new version of Eleventy parses strings in double curly brackets as template vars.

package.json
```diff
@@ -17,33 +17,32 @@
...
+        "fluid-binder": "2.0.0",
         "fluid-location-bar-relay": "1.0.7",
         "foundation-sites": "5.5.2",
-        "highlight.js": "11.4.0",
-        "infusion": "4.0.0-dev.20220209T141225Z.50bdbf8d1.main",
+        "highlight.js": "11.6.0",
+        "infusion": "4.5.0",
         "lunr": "2.3.9",
-        "markdown-it-anchor": "8.4.1",
-        "octicons": "3.1.0",
```
amb26, line 29: I'm not sure what the cause is, but I got an error running npm run build or npm run start complaining that markdown-it itself was not installed. 
````
[dev  ] [11ty] Eleventy CLI Fatal Error: (more in DEBUG output)
[dev  ] [11ty] 1. Error in your Eleventy config file 'E:/source/gits/infusion-docs/.eleventy.js'. You may need to run `npm install`. (via EleventyConfigError)
[dev  ] [11ty] 2. Cannot find module 'markdown-it'
````

Adding the line here:

        "markdown-it": "13.0.1",

made the problem go away

File: src/documents/Panels.md [hunk 5]
```diff
@@ -231,8 +231,8 @@ fluid.defaults("demo.panels.language", {
     protoTree: {
```
amb26, line 231: Not part of this work, but we could correct a typo here on line 63
````
Preference Maps are **only** required if you are working with
[schemas](PrimarySchemaForPreferencesFramework.md))
````
there is an extra ) after the link here

cindyli:
Good catch, @amb26. All addressed. This pull request is ready for another look. Thanks.

PR #263, opened 2024-04-24
feat: use pagefind for searching (resolves #262, #227, #153)

### LINKED ISSUE #153: Add button to search input
### screenshot: 
Screenshot from 2020-02-23 14-17-52

### additional context:
i would like to take this issue,

### LINKED ISSUE #227: Improve accessibility of the search fields on the Infusion Docs site
## Describe the bug
The search field is not properly labeled and isn't marked up as an input of type search.
https://github.com/fluid-project/infusion-docs/blob/05bba0b6e2509241b13d4b5c50656914b1e30314/src/layouts/default.html.handlebars#L94-L95

### Expected behavior

The search field should be marked up with proper accessible and semantic markup.

## Additional context or notes

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

Originally filed as [FLUID-6476](https://issues.fluidproject.org/browse/FLUID-6476)

### LINKED ISSUE #262: Update search to make use of Pagefind
## Is your feature request related to a problem?

Improve search results and indexing.

## Describe the solution you'd like

We are looking into using [Pagefind](https://pagefind.app) for our static sites. Pagefind provides fast indexing at build time, and small bandwidth usage. It is also used by 11ty's own site for its search. It should improve the search performance/results on this site and will be easier to maintain if the implementation is similar to our other sites as well.

jobara:
Resolves #262 
Resolves #227 
Resolves #153 

- Switched to using [Pagefind](https://pagefind.app) for searches
- Removed old search and related tests
- Updates implementation of mini search to just do a form submission and include a button
- Minor refactoring of styles related to contrast themes and updates related to Pagefind integration


greatislander (changes requested):
Links within search results lack sufficient contrast in the hover state:

Search results for the term 'UIO', showing 11 results for UIO, the first result under the heading Subcomponents has a link which appears with black text on a black background in its hover state

greatislander (changes requested):
It would be good to add a bit of space between the Pagefind "load more results" button and the footer.

greatislander (approved):

PR #266, opened 2024-04-29
GH-265: Restore 'octicon' link on heading hover (resolves #265)

### LINKED ISSUE #265: "Octicon" links not visible when hovering on a heading.
As @amb26 pointed out on the chats, the heading "octicon" icons no longer appear on hover.   I have confirmed that this is also the case when loading `main` locally.  I have confirmed that the markdown is still there, probably it's just a matter of adjusting styles.

When working on this we should also briefly confirm whether #242 is still an issue.

duhrer:
See #265 for details.


duhrer:
Note that I did not update to the much newer versions of the `octicons` library we used to use, as they really changed their approach a lot.

Also, I noticed that a lot of (now unused) search infrastructure from the old approach was there, and cleaned that up.

